PR #14, opened 2018-11-06
Tag selection fix closes #12

### LINKED ISSUE #12: Tag selection and deselection is faulty
Environment
-----------
Running Firefox 63.0.1 on Mac OS High Sierra.

Steps to reproduce (for bugs only)
-----------------------------
1. Run a version of timemaps with a hierarchy of tags. No events have been assigned tags.\
screen shot 2018-11-05 at 17 40 50

2. Open up a tag panel. In this case, the tag panel has three immediate children (but the bug seems pervasive across other tag setups).
screen shot 2018-11-05 at 17 42 11

3. Attempt to select the parent tag ('Soldiers').

Current Behavior
----------------
The box for the parent tag that has been selected is not highlighted to indicate that the tag filter is active. It does, however, seem to be selected 'behind the scenes', as if I close and open the tag panel or selected one of the children tags, the box for 'Soldiers' seems to become selected. 

Expected Behavior
-----------------
The box corresponding to the parent tag should be highlighted immediately when it is selected. This issue seems to exist in grandchildren tags as well: when I try to select a grandchild tag, it does not even become highlighted when closing and opening the panel.

Another bug that I noticed while testing this: if I select a tag that contains no events, it removes all events from the map (as expected). If I then deselect that tag, however, there remain no events on the screen. All events should return to the map after deselecting, as no selected tags should show all events.

francamps:
This pull request fixes issue #12, by simplifying event handlers passed to Checkbox and removing faulty code running in certain conditions when rendering tag tree after a check.

### DIFF
--- a/src/components/TagListPanel.jsx
+++ b/src/components/TagListPanel.jsx
@@ -21,26 +21,9 @@ class TagListPanel extends React.Component {
     this.computeTree(nextProps.tags.children[nextProps.tagType]);
   }
 
-  traverseNodeAndCheckIt(node, depth, active) {
-    // do something to node
-    const tagFilter = this.newTagFilters.find(tagFilter => tagFilter.key === node.key)
-    tagFilter.active = (depth === 0) ? !node.active : active;
-    tagFilter.depth = depth;
-    depth = depth + 1;
-
-    if (Object.keys(tagFilter.children).length > 0) {
-      Object.values(tagFilter.children).forEach((childNode) => {
-        this.traverseNodeAndCheckIt(childNode, depth, tagFilter.active);
-      });
-    }
-  }
-
   onClickCheckbox(tag) {
-    this.newTagFilters = this.props.tagFilters.slice(0);
-    let depth = 0;
-    if (tag.key && tag.children) this.traverseNodeAndCheckIt(tag, depth);
-
-    this.props.filter({ tags: this.newTagFilters });
+    tag.active = !tag.active
+    this.props.filter(tag);
   }
 
   createNodeComponent (node, depth) {
--- a/src/components/TagListPanel.jsx
+++ b/src/components/TagListPanel.jsx
@@ -30,7 +30,7 @@ class TagListPanel extends React.Component {
 
     if (Object.keys(tagFilter.children).length > 0) {
       Object.values(tagFilter.children).forEach((childNode) => {
-        this.traverseNodeAndCheckIt(childNode, depth, tagFilters, tagFilter.active);
+        this.traverseNodeAndCheckIt(childNode, depth, tagFilter.active);
       });
     }
   }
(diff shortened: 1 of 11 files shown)


francamps:
There is an implicit design decision here that is not resolved, and was implicitly coherent within the context of the Ayotzinapa case: the expectation that tags are unique (they were in Ayotzinapa).

Currently, tags have two functions: i) to characterize events by adding some consistent labels to them (tags are shown on Cards), ii) to filter in and out by checking boxes from the tag panels on the Toolbar, as well as searching through bundles of events grouped by those. 

A - If we decide to make individual tags being unique, I would say there should be a check that messages and enforces it.

B - If tags don't have to be unique, then the ways these tags are displayed as metadata for each event needs to be rethought, maybe by displaying each tag's entire lineage.

I don't think there is a right or wrong solution here, but rather a design choice that needs to be messaged properly.

src/components/TagListPanel.jsx
```diff
@@ -30,7 +30,7 @@ class TagListPanel extends React.Component {
 
     if (Object.keys(tagFilter.children).length > 0) {
       Object.values(tagFilter.children).forEach((childNode) => {
-        this.traverseNodeAndCheckIt(childNode, depth, tagFilters, tagFilter.active);
+        this.traverseNodeAndCheckIt(childNode, depth, tagFilter.active);
```
breezykermo: Is there a reason that we do this in the component? Shouldn't all validation be done in the reducer workflow?

src/reducers/utils/validators.js
```diff
@@ -18,6 +18,34 @@ function makeError(type, id, message) {
...
 
+
+const isLeaf = node => (Object.keys(node.children).length === 0);
+const isDuplicate = (node, set) => { return (set.has(node.key)); };
+
+
+/*
+* Traverse a tag tree and check its duplicates
+*/
+function traverseNodeAndCheckIt(node, parent, set, duplicates) {
```
breezykermo: maybe `isTagTreeValid`? (keeping with semantics of the other validation functions)

francamps: True, it needs a better name. How about `validateTree`? It doesn't return true/false, which is what I associate functions starting with 'is'.

breezykermo: Is there a reason it's not formatted as an 'is' function that returns true/false? I think I would prefer if we didn't remove duplicates, and instead just throw a descriptive error to say that there are errors in the tag tree.

src/components/TagListPanel.jsx
francamps: Yeah, you're probably right. I believe this logic (the `active` field per node should be determined in the selector.

src/reducers/utils/validators.js
breezykermo: Right -- I was suggesting that we enforce uniqueness by refusing to show anything of a tag tree that has duplicates, rather than modifying the tag tree before rendering; but thinking about it a little more I think I prefer this solution.

PR #17, opened 2018-11-10
Topic/basic smoke test

scztt:
A simple example of unit testing on travis - this should at least minimally validate that the server is launchable and can return some data. Usage of mocha is pretty minimal here, so this could easily be swapped for another framework if needed. This may or may not be the way ya'll would like to go re testing, but I think it's important to have at least SOME minimal green/red information on travis builds - this is a start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,4 @@ install:
   - yarn
 script:
   - yarn build
+  - yarn test
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "private": true,
   "scripts": {
     "dev": "webpack-dev-server --content-base static --mode development",
-    "build": "NODE_ENV=production webpack --mode production"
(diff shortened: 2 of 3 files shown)


scztt:
Hold off on merging this - travis has an issue with 
`subprocess.kill` that seems to be causing the test phase to time out. Looking in to alternative ways of managing the subprocess.

scztt:
Travis failures are resolved - this is now green.

breezykermo:
Thanks for this -- agreed, and looks good. My only request before merge would be to change the test runner to [Ava](https://github.com/avajs/ava), to keep the framework consistent between this repo and [datasheet-server](https://github.com/forensic-architecture/datasheet-server/).

scztt:
Yup, I noticed that - I started switching over to ava - one more problem to solve and then I'll add that.

PR #22, opened 2018-11-27
Feature: Narrative sequences closes #11

### LINKED ISSUE #11: Add support for narrative sequences of events
Adding support for displaying an event as logical followup to a previous event would allow to link events in logical and narrative sequences. This can be represented in a variety of ways, some of which are complementary:

* Events that are linked within a narrative sequence and follow each other would be visually bound by a directional edge.
* Selecting a time span in the timeline would filter events in the narrative sequence, displaying only those that fall within the displayed time bounds. This means at any given time zoom, TimeMap will only show those events in a narrative sequence that fall within those bounds, potentially leaving out some in the past or in the future.
* Cards with event metadata that, displayed on click, could show a 'Next'-like button that would let the user follow this sequence step by step.
* In the future, events that are linked within narrative sequences could show this in some form in the timeline also.

francamps:
This PR is a feature extension. It implements narratives, which as sequences of events threaded together. Narratives can be handy to describe trajectories as well simply sequences in time of events associated to each other.

Narratives can be individually styled via custom styles in the store, with the following fields:

```
narratives: {
  "narrative_1": {
          style: 'dotted',                  // ['dotted', 'solid']
          opacity: 0.4,                     // range between 0 and 1
          stroke: '#ffffff',                 // Any hex or rgb code
          strokeWidth: 2                 
  }   
}
```

The PR also contains a variety of code embellishments and cosmetic changes.


breezykermo:
Would you be able to update the 'example.config.js' in datasheet-server and [the wiki](https://github.com/forensic-architecture/timemap/wiki/Setting-up-a-local-instance-of-Timemap) if necesssary to reflect what is necessary on the backend for these changes?

breezykermo:
Also getting the following error on this branch: when setting up a Timemap and Datasheet Server instance as demonstrated in [this wiki](https://github.com/forensic-architecture/timemap/wiki/Setting-up-a-local-instance-of-Timemap), there seems to be an error when trying to use the timeline.

Clicking any button (arrows, or a scope option such as '3 days') gives the following error in console:

screen shot 2018-11-29 at 8 25 33 am

francamps:
@breezykermo As it is now, narratives have no metadata associated with them, and so they are not required to have a separate tab in the datasheet. That is, simply by adding a 'narrative' field to the events in question, the narrative feature is displayed. It comes with the same payload as events data. It can of course be empty, and nothing happens in that case.

If we add Narrative as a new object to fetch, I see it as the next phase of Narratives (part of a more 'Guided' mode of the application) and this PR will then extend significantly. I'm ok with doing that, but in that case, it may take a little longer.

breezykermo:
Ah, great, understood. Sorted that out on the example sheet and it works great! The one suggestion/question I'd have (but that I think can be sorted in a later PR so as not to interrupt development) is that it might make more sense to allow styling for color using SCSS variables, rather than JS, if possible... or otherwise to move other configurable styles like event color to the store also. Styling re-architecture should probably be a standalone PR though, as there's quite a lot of excess stuff from Ayotzinapa.

breezykermo:
Playing with this a bit more, there also seems to be a bug sometimes: when events are part of a narrative, I sometimes can't click on them to reveal their description etc. Maybe something to do with the narrative line being on top of the event SVG in the DOM?

breezykermo:
I tried rendering narrative lines beneath events, but the error persists. It seems to be something more systemic, to do with the way the handlers on event SVGs.
 * The `hash(app) !== hash(newApp)` check is not passing when an SVG in the map is clicked, meaning that `renderSelectedAndHighlight` is not called.
 * Even bypassing this check, the error persists.
I'm thinking the bug is something deeper to do with the way that state is tracked manually between 'map.js' and the Redux store. Note that it is possible to select an event through the timeline, which then makes it possible to toggle into some of the other description panels, but the dotted circle around the selected event remains fixed.

@francamps, could you look into this? More your domain than mine...

francamps:
It would seem that issue #28 is what explains this behavior. SVG event listeners seems to be working and the rendering of events, lines and highlight markers are in order as well. However, if the fetching of individual events fails on click, this error is not gracefully handled and so the application does not respond.

PR #29, opened 2018-12-03
Remove categoryGroups

breezykermo:
This PR removes `categoryGroups` from the frontend logic, a field that is no longer necessary given that categories now mean a different thing than they did in Ayotzinapa.

I have also cleaned up some of the React component logic, particularly to do with which components interpolate values directly from the Redux store. Previously, [Dashboard.jsx](/src/components/Dashboard.jsx) was passing down a bulk of values, which made debugging difficult. As a result, I have used react-redux `connect` in components such as `Viewport`, `CardStack` and `Timeline` to pare down the JSX of `Dashboard` (and in the proceess cleaned up some necessary props.


breezykermo:
[apologies for the interpolated git accounts -- i've fixed this on my git client for the future]

francamps:
I've added a small fix in CardStack and tried to continue the work a little further by removing the category group colors from `map.js` as well, as well as remove unused methods in Card. Overall looks like the good direction to be moving.

breezykermo:
Ah perfect, thanks for that fix!

PR #31, opened 2018-12-04
Fix/cleaner handlers

breezykermo:
* added a 'methods' argument when instantiating 'timeline.js', to keep at least a notionally consistent API between 'timeline.js' and 'map.js'. Methods were previously tacked inside the 'app' object.
 * fixed a bug where clicking on an event in the map and timeline were invoking the react handler with different arguments (one with ID list, the other with entire event list). Now both pass a list of events.
 * have renamed handler functions passed down from `<Timeline/>` so that it is clearer what they do when called in 'timeline.js'. All handlers that relay information to React/Redux level should be prefixed with 'on' and in camelCase, e.g. `onSelect`, `onToggle`.

### DIFF
--- a/src/components/CardStack.jsx
+++ b/src/components/CardStack.jsx
@@ -27,8 +27,8 @@ class CardStack extends React.Component {
             getCategoryGroup={this.props.getCategoryGroup}
             getCategoryColor={this.props.getCategoryColor}
             getCategoryLabel={this.props.getCategoryLabel}
-            highlight={this.props.highlight}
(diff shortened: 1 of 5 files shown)


breezykermo:
Note: we should probably rename the props passed through other 'portal' components in [Dashboard.jsx](/src/components/Dashboard.jsx) ('portal' here meaning a React component that manages some non-React process, e.g. Timeline->timeline.js and Viewport->map.js).

PR #39, opened 2018-12-06
Topic/guided narratives

francamps:
This PR implements Narratives as guided sequences of events. It adds them on the Toolbar, and displays a description of the current Narrative once it is being explored. The UI allows to travel from event to event on the narrative using the NarrativeCard, as well as the event Card being pulled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build/
 node_modules/
 config.js
+dev.config.js
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -212,3 +212,11 @@ export function toggleNotifications() {
         type: TOGGLE_NOTIFICATIONS
     }
 }
+
+export const TOGGLE_MAPVIEW = 'TOGGLE_MAPVIEW';
+export function toggleMapView(layer) {
+  return {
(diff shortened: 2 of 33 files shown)


breezykermo:
Has this been tested in the case that the `NARRATIVES_URL` returns a 404? I just tried to run the branch (without any narratives configured in the server), and it leaves me with no events showing in the map or the timeline. Errors seem to be coming from code to do with the narratives.

breezykermo:
On this note, it would probably be good to implement better error relaying from the `timeline.js`/`map.js` layers up through the React layer. Or, at least, make them throw noisy errors in the console. I've been finding when trying to debug at that level of the application that there can be nothing on screen, and a silent fail, which isn't very heartening

francamps:
Handled the error when NARRATIVE_URL returns an error, and rebase the current state of develop.

breezykermo:
Looks great! Just removed a stray `console.log`, and have added a note in the README regarding the required URL. The only thing it's missing is an update in [the wiki](https://github.com/forensic-architecture/timemap/wiki/Setting-up-a-local-instance-of-Timemap), which I'll do now, and then merge.

Issue #28, opened 2018-11-30
Graceful error handling for failing individual events fetching


francamps:
Current Behavior
----------------

In the current state of Timemap, when the user clicks on an event on the map, the application makes a single fetch to the server for additional metadata (such as description), and the response is used to populate the Cards on Cardstack.

However, if the fetch fails the error is not properly handled. This can cause two different situations:

* If the cardstack was already open with some other event, but the new event fails to fetch, the card displays the loading spinner and no message is displayed, except for an error in the console.
* If the cardstack was not open, nothing happens, except for an error in the console.

Expected Behavior
-----------------

The user should be notified that the fetching of that specific event failed. 

I suggest using the Notification component to display the error, and creating a default empty card as well.

I've noticed this in the `feature-narrative-seq` branch, but it should be easy to reproduce in the current state of `develop`.

#### Steps to reproduce

1. Use your instance of TimeMap normally and focus on a point in the map where some events are display. 
2. Stop the data server **without** reloading the page. 
3. Now click on any of the events.

PR #41, opened 2018-12-10
Topic/redesign card

francamps:
This PR mostly restructures React components for Card and implements aestethic changes to its design.


breezykermo:
works nicely for me. Not sure I'm a fan of this shadowing:

screen shot 2018-12-10 at 15 17 40

Is there a logic to this? Happy to merge though. I'm further redesigning the experience of these cards with sources in #28, so not fussed about getting it perfect here.

breezykermo:
Also a question: what happens to 'Connected Events' if the current event is a waypoint in more than one narrative? How do we currently prioritise which narrative is displayed?

Issue #20, opened 2018-11-15
Support sources -- media associated with events


breezykermo:
The atomic concept in Timemap is the `event`, which represents a distinct moment in time and/or space. Each event has a range of characteristics such as; date, time, a description, and so on. 

When working on investigations, it is important that we credit events with the source or sources from which they were derived. To this end, there needs to be a `source` concept. Each source can in theory consist of images, videos, audio, and/or text. There is a many-to-many relationship between sources and events, as just as an event could have multiple sources, so too could a source be referenced by many events.

A proof-of-concept for a sources should meet several other criteria:
 * Sources need to be created dynamically by externally served data, i.e. there needs to be a 'sources' endpoint.
 * Sources should contain _references_ (URLs) to media, rather than the media themselves, as they will be loading on demand (when a user clicks an event to view its contents). Text fields can be included directly in sources.
 * There does not need to be a direct visualisation of sources in the proof-of-concept, but we should keep in mind the possibility of visualising sources independently in the future.

francamps:
Originally, TimeMap had two somewhat similar concepts that are done away with in the current version. There was a notion of 'Account' which was a person or entity according to who this event is accounted for. 

Additionally, there was a notion of 'Source', which strictly referred to the material origin of this information. This means that for example if we know of an Event through the testimony of a witness (W) compiled in a report R, Account would refer to W, while Source would refer to R. 

This solution, although implemented with problems, allowed to group by accounts or testimonies, as well as specific pieces of media where information was compiled. Accounts or testimonies can be now accounted by categories and category groups, but will probably need more thinking. This is not the purpose of this issue though.

Now, this current form of source is different and I like it more because it allows multiple sources per event, which could then ideally be compared on a per event basis. 

Does the following make sense for Source's?

```
{
    id
    title
    description
    type
    link
}
```

breezykermo:
Here's what I was thinking regarding the relationship between sources and events.

There is a many-to-many relationship between events and sources: each event has 1 or more sources, and likewise, a source can be used across multiple events.

Within each source, you can have a range of different blocks that compose the source:
 * captioned image
 * captioned video
 * hypertext (link to a testimony source, or other)

This then leads to the right-hand panel that appears when you click on an event in Timemap being composed of the following:
 * event description, time, location, and other metadata
 * clickable 'prev' and 'next' arrows that take you to the previous or next events in each narrative of which the event is part.
 * a list of sources for the event, where each source has associated media as described above.

[...]

PR #44, opened 2018-12-12
Topic/source fetch

### LINKED ISSUE #28: Graceful error handling for failing individual events fetching

breezykermo:
closes #28, and prepares for #20.

* the error dispatching in a `FETCH_SOURCE_ERROR` action is now displayed in Card.
* loading is showed appropriately in Card, only in the 'source' field.
* cards are expanded by default, in preparation for a more continuous scroll through multiple events

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -118,7 +118,6 @@ export function updateDomain(domain) {
 
 export function fetchSelected(selected) {
   if (!selected || !selected.length || selected.length === 0) {
-    console.log('hitting base')
     return updateSelected([])
   }
   return dispatch => {
@@ -127,6 +126,7 @@ export function fetchSelected(selected) {
       dispatch(fetchSourceError('No source extension specified.'))
     } else {
       dispatch(toggleFetchingSources())
+      // TODO: fetching logic
     }
 
   }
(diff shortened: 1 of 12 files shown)


breezykermo:
Right. I guess that I would prefer 'highlighting' to not entail a visual unfolding of the card, as it obscures that detail for me unnecessarily. Would you be open to highlighting being shown in some other way, e.g. foregrounded with opacity 1, whereas other cards have opacity 0.7?

Issue #40, opened 2018-12-10
More descriptive error-handling when updateDomain fails


breezykermo:
A badly engineered `UPDATE_DOMAIN` action fails silently. This can be thought together with #28; the error architecture needs to be more elegant in general.

PR #52, opened 2018-12-17
Implement focused narrative mode

### LINKED ISSUE #42: design narrative mode
A user can enter into narrative mode, either by selecting one of the narratives from the left menu, or by clicking the 'next' or 'previous' button when an event is selected in explore mode.

This mode needs to feel slightly different than explore mode:
 * a slight shadow around the perimeter of the viewport
 * events that are not a part of the narrative dimmed
 * trajectories of other narratives disappeared, current narrative trajectory foregrounded.
 * some way to exit narrative mode, perhaps by pressing a 'return to explore', or by cancelling the narrative panel in the top left (which currently cannot be closed).

### LINKED ISSUE #43: Narrative lines display when events are not shown
When the timeline is narrowed to filter events (or when a tag is specified), the events disappear from the map:  but the dotted lines that string events together to represent narratives do not.

@francamps, I think that we need to do some more design thinking on narratives; in particular, how they are (or are not) visualised in the map.

I am beginning to think that maybe narratives should not be represented as dotted lines in the map, but that dotted lines (or other visual connections between events) are activated by some other data structure. Not sure what the solution is here, but flagging to start a discussion.

francamps:
This PR implements a more focused narrative mode, closes #42 and closes #43. 
The user can now select a Narrative in the toolbar. Selecting a narrative:
* Foregrounds a curated sequence of events that are bundled together thematically around an articulated narrative.
* Hides the toolbar to help the user focus on that sequence.
* Displays a Narrative card that summarizes the sequence and allows to move from event to event.
* Displays a styled sequence of arrows on the map from event to event in the sequence.
* Displays a toned down sequence of arrows associated with other available narratives. 
* When closing the narrative card, the user goes back to simple exploratory mode: the Toolbar displays again, and narrative arrow lines become hidden.

### DIFF
--- a/src/components/Timeline.jsx
+++ b/src/components/Timeline.jsx
@@ -47,6 +47,7 @@ class Timeline extends React.Component {
     const labels_title_lang = copy[this.props.app.language].timeline.labels_title;
     const info_lang = copy[this.props.app.language].timeline.info;
     let classes = `timeline-wrapper ${(this.state.isFolded) ? ' folded' : ''}`;
+    classes += this.props.app.narrative ? 'narrative-mode' : '';
     const date0 = formatterWithYear(this.props.app.timerange[0]);
     const date1 = formatterWithYear(this.props.app.timerange[1]);
 
@@ -80,7 +81,8 @@ function mapStateToProps(state) {
       timerange: selectors.getTimeRange(state),
       selected: state.app.selected,
       language: state.app.language,
-      zoomLevels: state.app.zoomLevels
+      zoomLevels: state.app.zoomLevels,
+      narrative: state.app.narrative
     },
     dom: state.ui.dom,
   }
--- a/src/components/Timeline.jsx
+++ b/src/components/Timeline.jsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import * as selectors from '../selectors';
+import hash from 'object-hash';
 
 import copy from '../js/data/copy.json';
-import { formatterWithYear } from '../js/utilities';
+import { formatterWithYear, isNotNullNorUndefined } from '../js/utilities';
 import TimelineHeader from './presentational/TimelineHeader';
 import TimelineLogic from '../js/timeline/timeline.js';
 
@@ -21,7 +22,9 @@ class Timeline extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.timeline.update(nextProps.domain, nextProps.app);
+    if (hash(nextProps) !== hash(this.props)) {
+      this.timeline.update(nextProps.domain, nextProps.app);
+    }
   }
 
   onClickArrow() {
@@ -32,7 +35,7 @@ class Timeline extends React.Component {
 
   render() {
     let classes = `timeline-wrapper ${(this.state.isFolded) ? ' folded' : ''}`;
-
+    classes += (this.props.app.narrative !== null) ? ' narrative-mode' : '';
     return (
       <div className={classes}>
         <TimelineHeader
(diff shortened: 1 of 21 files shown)


breezykermo:
This is excellent stuff. There are only a few bugs I have from some basic user testing, plus quite a few questions that I'm listing here to keep everything together.

I'll also review your implementation more closely using GH review to comment on the architecture of the feature.

## Issues
 * There is a palpable delay when moving into narrative mode with a larger number of narratives (I have 25 in a demo sheet). Can we add some sort of loading modal? Am also curious as to why this delay is necessary; is there a long-running/async process required in moving to narrative mode?
 * _corner case_; when there is only one event in a narrative and we move into narrative mode, all events disappear and there is a single red triangle displayed. One solution here could be to disqualify narratives that only have one event when they are loaded from a remote.

## Questions
 * The narrative seems to be displayed with an arrowhead on the first event in the narrative, as well as an arrowhead on the last. Would it make sense to have an arrowhead at every event?
 * I see the other narratives as dark grey lines, which is lost against the grey-out mapbox tiles at a certain zoom. Could we maybe make them a less opaque version of whatever colour the narratives are (e.g., yellow)?
 * How difficult will it be to make less opaque the events _not_ in the current narrative? We can definitely put this as a feature after this branch is merged, but I think it is am important feature, so that the location of the events in the current narrative are distinctly foregrounded and not muddled with other close-by or co-located events.
 * Is there any way to dynamically move between narratives in narrative mode? Again, a feature we can put in a supplementary issue. I think it would be great to be able to double click on the faint lines of narratives to switch, or even offer more dynamic path-making at junction events.

breezykermo:
I also tacked on a basic commit that is a placeholder for #40, not explicitly to do with narrative mode but is an important bug to fix.

breezykermo:
Playing around with this a little more; the bug where the mapbox tiles give way to white at a certain zoom resolution has returned for me in this branch. Was there a previous commit that fixed it, or am I imagining things and it's really always been there?

breezykermo:
Have added the performance concern identified here as #53, otherwise looks good to me!

src/components/Timeline.jsx
```diff
@@ -15,18 +17,14 @@ class Timeline extends React.Component {
...
-    this.timeline = new TimelineLogic(this.props.app, ui, this.props.methods);
+    this.timeline = new TimelineLogic(this.props.app, this.props.ui, this.props.methods);
     this.timeline.update(this.props.domain, this.props.app);
-    this.timeline.render(this.props.domain);
   }
 
   componentWillReceiveProps(nextProps) {
-    this.timeline.update(nextProps.domain, nextProps.app);
-    this.timeline.render(nextProps.domain);
+    if (hash(nextProps) !== hash(this.props)) {
```
breezykermo: Is this hash necessary? I understand it is a way of doing deep prop comparison, but potentially it is contributing to the performance concern?

src/js/map/map.js
```diff
@@ -360,61 +369,132 @@ Stop and start the development process in terminal after you have added your tok
...
-        const styleProps = getNarrativeStyle(d[0].narrative);
-        return styleProps.stroke;
-      })
-      .style('fill', 'none');
+
+    narrativesDom.selectAll('.narrative')
+      .each((g, i, nodes) => { return updateNarrativeSteps(g, i, nodes); });
   }
 
+  function updateNarrativeSteps(g, i, nodes) {
```
breezykermo: Wowee, lots of JS SVG logic. I imagine that _this_ is what's causing performance concerns. I think we ought to incrementally refactor this logic into SVG markup rendered by React components, as described in #53

PR #54, opened 2018-12-19
Break down map logic into React components and SVG markup

francamps:
This PR is a first pass at breaking down the previous map logic, which was a combination of vanilla js, d3 and Leaflet, into a series of standalone React components that compute mapping properties and render svg directly. Leaflet still offers displays tiles and provides projection calculations and animations to movement and zoom, but React is now carrying most of the element rendering load, which includes sites, events, narratives and markers.


breezykermo:
Wow -- moving around the application feels so much better. Would you be able to explain a little more of the architectural changes in this PR? For example, I see that react-portal is added: what is this dependency doing? Can you explain how exactly the code now leverages React caching better?

What do you mean when you say "React is not carrying most of the element rendering load"? Isn't React now in fact responsiblle for _more_ of the decision making regarding whether or not an element is rendered?

PR #58, opened 2018-12-26
Topic/source async

### LINKED ISSUE #20: Support sources -- media associated with events

breezykermo:
closes #20 

This PR adds basic source support in the application. Sources are broadly defined as discussed in #20. The metadata for sources is loaded up front (along with events, narratives, etc), and then additional media data such as photos, videos, or documents are loaded dynamically when the source is selected.

As per the modified [example.config.js](https://github.com/forensic-architecture/timemap/blob/topic/source-async/example.config.js), the metadata for sources are loaded into the application as single object, where the key is a unique ID to identify the source, and the value is a source object. The source object should be structured as follows:

```js
{
  id: 'The Name of the Photo', // ID is currently both the identifier, and the main title for the source.
  type: 'Photo', // currently supports one of ['Photo', 'Video', 'Eyewitness Testimony']
  path: 'https//some.url/path-to-photo',
  url: 'https://another.url/path-to-original', // optional URL to original source
  title: 'Some other string', // optional secondary title
  date: '12/03/1994' // optional date to display
}
```

This structure is very WIP and subject to change in future PRs. The format for a source's `path` has not been well standardised, and the solution is currently a load of hacky code that caters specifically to our file extensions and source media structure. I am using [gohttpserver](https://github.com/codeskyblue/gohttpserver) in development to serve media from a local URL. (The localhost URL is prefixed using Google Sheet expressions _in the datasheet source_).

Design for the source modal could also use some work. The primary reason for PRing this now is due to a looming internal deadline for a feature set of Timemap in mid January.


breezykermo:
I've just updated the [example sheet](https://docs.google.com/spreadsheets/d/1UC7DkCFeUXHfpUxUGruExwFbP4pqVBdJLOKfo6wDDGk/edit#gid=974355827) appropriately for this.

There are a few issues in the example sheet. In order to stop it from dragging behind, I think we need to require an example instance, working from the setup in the example configs, before merging into develop. It would be great to automate this process using Travis.

breezykermo:
To test with the example sheet, download the following folder:
[https://drive.google.com/drive/folders/1uLJuIiyX0yTYWYFlgttalUf8OkxPJj5U?usp=sharing]()

Terminal into the folder, and run: `docker run -it --rm -p 8000:8000 -v $PWD:/app/public --name gohttpserver codeskyblue/gohttpserver`.

If you have Docker installed, this should serve the assets from a server at [localhost:8000]() that will work with the paths in the example sheet (which are specified with the localhost prefix [in export](https://docs.google.com/spreadsheets/d/1UC7DkCFeUXHfpUxUGruExwFbP4pqVBdJLOKfo6wDDGk/edit#gid=1584139420).

Make sure also that you pull from the develop branch of datasheet server, and run using [the new sheets-config](https://github.com/forensic-architecture/datasheet-server/blob/develop/src/sheets_config.js).

breezykermo:
The feature needs some styling: but it now serves photos, photobooks, and videos. There is not yet support for Markdown, but I will add this in a subsequent PR.

example.config.js
```diff
@@ -1,18 +1,20 @@
...
-  title: 'Example',
+  title: 'example',
   SERVER_ROOT: 'http://localhost:4040',
-  EVENT_EXT: '/api/example/export_events/rows',
+  EVENT_EXT: '/api/example/export_events/deeprows',
   CATEGORY_EXT: '/api/example/export_categories/rows',
-  SOURCES_EXT: '/api/example/export_events/ids',
-  NARRATIVE_EXT: '/api/example/export_narratives/ids',
+  NARRATIVE_EXT: '/api/example/export_narratives/rows',
+  SOURCES_EXT: '/api/example/export_sources/ids',
```
francamps: This url seems to work with `/deepids` not `/ids`, so this file should reflext that.

src/components/presentational/NoSource.js
```diff
@@ -0,0 +1,22 @@
...
+const NoSource = ({ failedUrls  }) => {
+  return (
+    <div className="no-source-container">
+      <div className="no-source-row">
+        <i className="material-icons no-source-icon">
+          error
+        </i>
+        <div>No media found, as the original media has not yet been uploaded to the platform.</div>
+      </div>
+      {/* <div className="no-source-row"> */}
```
francamps: I understand this to be the preferred mode for this component but not having been fully fleshed out yet.

example.config.js
breezykermo: Thanks, corrected in 2nd commit below.

src/components/presentational/NoSource.js
breezykermo: Correct; I've removed these comments for the time being and am including it in an issue for better styling as suggested.

Issue #60, opened 2019-01-02
More coherent/consistent styling in SourceOverlay


breezykermo:
#58 added all necessary functionality for sources, but the overlay is not the prettiest at present.

<img width="640" alt="screen shot 2019-01-02 at 15 48 03" src="https://user-images.githubusercontent.com/16495856/50599375-d3f5ba00-0ea5-11e9-8cbe-858feac5ef91.png">

The styling needs to be more elegant, and more consistent with the rest of the application. In particular:
 * Failed images and videos should provide a descriptive error message that states which URLs were tried for the asset.
 * A coherent design language for sources composed of an abstract number of images (in 'Photobook' sources) implemented.
 * The outline of the modal needs to be responsive, both to the side of the window, and the size of the media it contains.

PR #66, opened 2019-01-03
Add support for markdown files in sources

### LINKED ISSUE #61: Sources should not be tightly coupled to source types
The layout of the media displayed in SourceOverlay, as well as the default icon displayed in event Cards are both currently determined by [hardcoded source types](https://github.com/forensic-architecture/timemap/blob/topic/source-async/src/components/SourceOverlay.jsx#L74-L84).

This is inflexible, and does not accommodate sources that are composed of multiple types of media. Sources need to be able to contain (and display) an arbitrary number of images, videos, and potentially other types of media. The default source icon does not need to be specific; instead the thumbnail image should be presented, or some reasonable default thumbnail/icon displayed when there is no thumbnail specified.

### LINKED ISSUE #62: Sources should support Markdown text
With the adjustment to source structure in #61, we also need to support Markdown files as supplementary media related to a source. This will allow full flexibility in flexibly composing a coherent multimedia source from images, video, and text.

Display of the media in sources should be determined by the file extension provided in the path. A reasonable error should be displayed when a path includes an unsupported file extension.

breezykermo:
closes #61, closes #62 

Source content can now be composed of an arbitrary collection of images, videos, and markdown files.

This PR decouples the display of sources from strictly support source types (such as 'Photo', 'Photobook', 'Video', and so on), and instead dynamically displays the contents in a source's paths based on the file extensions of the URLs. For example, if a source existed of one photo and one markdown file, the source is displayed:

screen shot 2019-01-03 at 13 11 24

The PR also:
* adds better error handling for thumbnails, by defaulting to an icon if neither a thumbnail is provided, nor any images exists in the source contents.
* improves the styling and responsiveness of the source modal slightly (still work to go here, so not closing #60).


lostground:
Using the `develop` branch of timemap and the `deepids` blueprinter I am trying to include a number of Markdown files as sources for events. The files are hosted on port `localhost:8000` fine, however the text does not render in timemap if I click on the source. Just the frame is loading, but with no content. No error is being displayed either.

Don't know if it's related but navigating to `localhost:8000/mardownfile.md` results in the .md file being downloaded rather than rendered in the browser. I've tried to use the Markdown Viewer browser extension, but doesn't help.

.png and .mp4 files render and play fine in the browser both in timemap and on a standalone page, just .md isn't working on my machine.

breezykermo:
Is timemap requesting the correct URLs? You can open the Network tab in Chrome/Firefox dev tools and see whether Timemap is correcting the request file.

Another issue that we have encountered in datasheet-server development  is CORS errors, is this one of those issues?

lostground:
> Another issue that we have encountered in datasheet-server development is CORS errors, is this one of those issues?

It looks like it's this.

I get this error in Firefox, when trying to load the sources tab. This is the only error in my console.
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://localhost:8000/123.md. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).
```

> You can open the Network tab in Chrome/Firefox dev tools and see whether Timemap is correcting the request file. 

Did you mean requesting the correct file? ;) 

Yeah, I'm pretty sure. My content server is returning `127.0.0.1 - - [14/May/2019 15:24:12] "GET /123.md HTTP/1.1" 200 -' so it looks like it's being queried correctly.

breezykermo:
Yes, this is a CORS issue. How are you running the content server (that serves the markdown)? You need to make sure that CORS is enabled on that server.

For example, you can use [http-server](https://www.npmjs.com/package/http-server) and run:

`http-server --cors -p 8080`

lostground:
Yes, thanks a lot! I was running the content server just with python SimpleHTTPServer before. Using npm's http-server with the code you gave me solved the problem.

Issue #53, opened 2018-12-18
All SVGs rendered on domain/app update


breezykermo:
There is a lot that could be improved about the handoff from the React component hierarchy to the `map.js` and `timeline.js` classes, which implement d3 logic using selectors.

These classes current have initialisation logic that is called in the `componentDidMount` of proxy React wrapper components, as well as `render` and `update` functions that are called conditionally in the `componentWillReceiveProps` method.

The problem with this is that we lose the benefit of React's [reconciliation](https://reactjs.org/docs/reconciliation.html) algorithm when the Redux store updates, requiring **all** SVGs to be re-rendered in both the timeline and the map on any change.

The fix for this problem will require a relatively hefty refactor of the d3 logic in those two classes, to better make use of the caching we get when using React components. Essentially, what is needed is _a refactor of timeline.js and map.js into a range of React components that render the equivalent SVG markup_.

francamps:
I have been thinking that this is a good idea and we can work on it progressively. It will be an interesting route to go, and there is some low hanging fruit on reactifying parts of map.js and timeline.js. I would say based on the discussion that map.js is more urgent than timeline.js since it operates on svg via js a lot more, and it generally contains more elements.

I agree that leveraging the reconociliation benefits to avoid rerenders can help us. There will be some things that we might want Leaflet (or Mapbox GL) and d3 to handle, in terms of scales, transforms and animations, but we can start for example with the svg itself and the site labels and go from there.

PR #68, opened 2019-01-04
Topic/reactify timeline

francamps:
This PR breaks down the previous d3-based rendering of Timeline in React components, which are now, for the most part, responsible for the rendering of the timeline. 

The top component, `Timeline`, has a scale in the x axis (a time scale) and one in the Y axis. All the operations over time happen in this component, and some children of it refer to the parent scale operations so that they render in sync.

d3 keeps the responsibility of the time scale on the x axis, in terms of computing positions of events and markers. The only rendering responsibility d3 keeps is the time axis and its ticks. The track where events are rendered stays _draggable_, meaning the user can still drag the time backwards and forwards, and this behavior is still determined by d3 convenient drag event handler.

However, React is now in charge of rendering and animating events, markers, categories in the y axis, labels and controls. 

The PR includes some very minor style changes.


francamps:
Closes #53

breezykermo:
A couple of preliminary notes just playing around:

The left/right arrows to move the timeline range give an error when clicked in narrative mode. I'm okay with just removing the arrows. I've also been thinking that it would be nice if we could come up with a clever way to adjust the scope of both the timeline and the map when entering a narrative based on what events that narrative contains... I'll open up a separate issue for discussion.

I love the animation that strings the selected halo between events. The only thing is that it is quite fickle for me: it doesn't quite seem to know how to re-render out past frames.
screen shot 2019-01-04 at 17 33 30
Is there a way that we can make the animation feel a little more reliable? Am not sure if this is intentional or not.

src/components/Timeline.jsx
```diff
@@ -4,27 +4,106 @@ import * as selectors from '../selectors';
...
+        margin_left: 120,
+        margin_top: 20,
+        trackHeight: 80      
+      },
+      softTimeUpdate: 0,
+      scaleX: null,
+      scaleY: null,
+      timerange: [null, null],
+      dragPos0: null,
+      transitionDuration: 300
```
breezykermo: Having just re-factored the state for narrative mode out of NarrativeCard in #67, I wonder whether it makes sense to put most of this in the redux state, rather than keeping it in component here. I realise that there is some logic that may need to stay here as state to play nicely with the d3 methods, but I think that you could get rid of some of the logic in both `componentWillReceiveProps` and `componentDidUpdate` by moving some values to redux.

src/components/TimelineEvents.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+class TimelineEvents extends React.Component {
```
breezykermo: This can be a stateless component.

src/components/TimelineMarkers.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+class TimelineMarkers extends React.Component {
```
breezykermo: Also can be a stateless component.

src/components/TimelineZoomControls.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+class TimelineZoomControls extends React.Component {
```
breezykermo: Also should become a stateless component.

francamps:
I removed the animation for the marker that was bugging out in Firefox, and moved a bunch of these components to stateless ones. 
I removed one unused state property from the local Timeline, but I think I'd rather keep the others (`scale`s, `dragPos0` and `timerange`) in the local state because they are mostly inner workings of the timeline. For example, in the case of timerange, it is kept in the local state so that it doesn't trigger a fresh reselect of all the data.

breezykermo:
I would definitely like to consider moving `timerange` to redux state, as the fact that it is currently being ferried to state in `componentWillReceiveProps` and also tailed with a call to `onUpdateTimerange` in `onMoveTime` indicates for me that it is data that resides in redux state, rather than component-specific state.

Even if this necessitates a fresh reselect, I think that it would be beneficial for the app architecture in general, as I doubt that the reselect would actually be a performance issue, as React will still deal with whether or not to re-render the component efficiently.

francamps:
So the reason this local state `timerange` exists at all is so that it can keep a momentary variation of timerange in the local state every time the drag event updates _during dragging_. That is, on every frame while dragging, the local timerange updates, and when releasing, the redux timerange is updated, by consequence this local one as well. Its name could be changed to make that more clear, to `dTimerange`, or `localTimerange` or something. 
This is only necessary for dragging, as onMoveTime or onApplyZoom are single actions, and so can and must afford to trigger a redux state update. Every time there is a call to `onUpdateTimeRange` as action, the redux state timerange is updated.
When there is the amount of data in order of Ayotzinapa (which is not even _that_ much, but a few thousand or tens of thousand rows), for example, it can get noticeable. I don't see it as a big deal if you think of it as the inner mechanics of the dragging behavior.

breezykermo:
Ah I see now -- just had a play with the code and came to realise what was happening, then saw your comment above.

Are the presets for `this.state.dims` hardcoded at the component level? Should they be derived from the timerange in redux state?

breezykermo:
Oh one other thing to keep consistent here @francamps: could we also dim the opacity of events that are not in the active narrative in narrative mode, to keep consistent with events in map?

PR #69, opened 2019-01-04
Step styles

breezykermo:
Enables optional styling per step in narratives, rather than requiring that one step style dictate all steps in a given narrative.

If `NARRATIVE_STEP_STYLES` is enabled in config.js, the application looks to `state.ui.narrative.stepStyles` for the style per step in each narrative.

The style per step is specified in each event. Along with the `narratives` array, there is an equal length array called `narrative___stepStyles` that contains the titles of the styles that should be drawn for the step from the given event to the next in each narrative. `narrative__stepStyles` needs to be an array, as the step from one event to the next may be styled differently across narratives (if the event exists in multiple narratives).

`narrative___stepStyles` is not a great name. In future PRs, we should move towards decoupling attribute names coming from the server and items on data structures in Timemap as much as possible, so that the name of the attribute in the event object coming from the server does not need to directly match the name used in the frontend, but rather can be configured if necessary.

### DIFF
--- a/src/components/MapNarratives.jsx
+++ b/src/components/MapNarratives.jsx
@@ -18,17 +18,16 @@ class MapNarratives extends React.Component {
     return this.props.narrativeProps[styleName]
   }
 
-  getStrokeOpacity(narrative, step) {
-    if (this.props.narrative === null) return 0
-    if (!step || narrative.id !== this.props.narrative.id) return 0.1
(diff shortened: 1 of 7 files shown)


breezykermo:
There is also included a fix here for proper error reporting when there is a bad path in text sources, which was missed in #62

PR #78, opened 2019-01-08
Topic/refactor event styles

### LINKED ISSUE #71: decouple event sizing in explore mode from quantity
The old design of events in explore mode caused the event SVG to become larger when multiple events were displayed in the same location. 

Visual effects like enlargement should be decoupled from data inputs. Instead, the effects should be configurable, so that different data logics can drive different effects across Timemap instances.

What this means practicably is essentially allowing a custom function to be passed down to `MapEvents` that is invoked during `renderLocation`, so that certain attributes of the SVG or element can be configured.

We should also add support for animations when event SVGs are mounted/unmounted. React's idiom for animations uses state, which we need to reimplement to recover then functionality we lost when we reactified the map and timeline components in #54.

breezykermo:
This PR creates two functions `styleLocation` and `styleDatetime` in `Map` and `Timeline` respectively that allow additional styles to be composed for the SVGs that are rendered in those components. 

Locations and datetimes are intermediate representations by which events become visual in the component. (Locations in `Map` existed before this PR, datetimes in `Timeline` were added.) A datetime is the vehicle that ferries one or more events (all those that occur at the datetime) from the redux domain to a visual representation on the timeline. Each datetime is represented as a single circle on the timeline by default. Locations in `Map` work the same way.

The functions `styleLocation` and `styleDatetime` are props passed to the `MapEvents` and `TimelineEvents` components that allow custom style props and render items to be added to the SVG location/datetime. This is useful, as it means that custom representations of events in timeline/map can easily be written as a function of the domain, or any other part of the redux state.

By default, all locations in `Map` are now the same size. If there is more than one event contained by a location, a `<text>` element with the appropriate count is added to the rendered representation in `Map`.

closes #71

### DIFF
--- a/src/components/Timeline.jsx
+++ b/src/components/Timeline.jsx
@@ -221,9 +221,7 @@ class Timeline extends React.Component {
    * components in the <g/> div.
    */
   styleDatetime(timestamp) {
-    return [
-      { fill: 'orange' },
-    ]
+    return []
   }
 
   render() {
--- a/src/components/Timeline.jsx
+++ b/src/components/Timeline.jsx
@@ -212,9 +212,10 @@ class Timeline extends React.Component {
   }
 
   styleDatetime(timestamp) {
-    return {
-      fill: 'orange'
-    }
+    return [
+      { fill: 'orange' },
+      () => <text>ciao</text>
+    ]
   }
 
   render() {
--- a/src/components/Timeline.jsx
+++ b/src/components/Timeline.jsx
@@ -19,6 +19,7 @@ class Timeline extends React.Component {
   constructor(props) {
     super(props);
     this.styleDatetime = this.styleDatetime.bind(this)
+    this.getDatetimeX = this.getDatetimeX.bind(this)
     this.svgRef = React.createRef()
     this.state = {
       isFolded: false,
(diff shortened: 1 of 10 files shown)


francamps:
I like the idea and it works particularly well on map. However, we are losing the capacity to see the presence of events of multiple categories in one location, and the color that ends up being displayed is arbitrary, simply the last one on top. Maybe the default function for styling could add a little randomization of the X position, so you could see a stack of circles of different colors? As an idea.
Also, would it be possible to make the number-in-event-circle text white by default? I think it could help legibility.

In the timeline, the datetime object need to account for the distinction by category, and I can't find if it's doing it or not. In practice, I'm getting all events in the same track of the timeline like this:

screen shot 2019-01-09 at 10 57 01 am

src/components/presentational/DatetimeDot.js
```diff
@@ -0,0 +1,27 @@
...
+  events,
+  x,
+  y,
+  styleProps,
+  extraRender
+}) => (
+  <g
+    className='datetime'
+    transform={`translate(${x}, ${y})`}
+    onClick={() => onSelect(datetime.events)}
```
francamps: It looks like `onSelect` here is undefined.

francamps: also `datetime.events` should be just `events`

src/components/Timeline.jsx
```diff
@@ -282,7 +234,57 @@ class Timeline extends React.Component {
...
+                dims={dims}
+                onApplyZoom={(zoom) => { this.onApplyZoom(zoom); }}
+              />
+              <TimelineLabels
+                dims={dims}
+                timelabels={this.state.timerange}
+              />
+              <TimelineMarkers
+                selected={this.props.app.selected}
+                getEventX={(e) => this.getEventX(e)}
```
francamps: `getEventX()` should change to `getDatetimeX()`

src/components/Timeline.jsx
```diff
@@ -282,7 +234,57 @@ class Timeline extends React.Component {
...
+                getEventY={(e) => this.getEventY(e)}
```
francamps: Probably should change to `getCategoryY()`

PR #80, opened 2019-01-08
First zoom narrative

### LINKED ISSUE #72: If event is selected, timeline zooms should keep it in focus
When an event is selected and the timeline zoom is intensified, the selected event does not stay in focus (unless it happens to be exactly in the middle of the timeline). Zoom modifications of the timeline should gravitate around selected events, keeping them in focus.

### LINKED ISSUE #73: Adjust zoom of timeline and map appropriately when moving to narrative
When entering a narrative, the zooms of the map and timeline respectively need to be adjusted to some level that is appropriate to showcase all or most of the events in the narrative.

To calculate the zoom for the map, for example, a conceptual bounding box should be drawn around all of the events in the narrative, and the screen size moved to this bounding box.

Likewise for the timeline.

francamps:
This PR implements a series of zooming features to improve the experience.
When the user enters narrative mode by selecting a narrative in the toolbar:
- The timeline zooms to a scope that spans all of the events in the narrative, by zooming in or out accordingly.
- The map zooms in or out to a level where all events in the narratives are visible.
This closes #73.

Additionally, when the user selects an event, the timeline automatically now centers itself on the selected event, without changing the zoom. this closes #72.

### DIFF
--- a/src/components/Map.jsx
+++ b/src/components/Map.jsx
@@ -35,17 +35,19 @@ class Map extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (hash(nextProps.app.selected) !== hash(this.props.app.selected)) {
-
-      // Fly to first  of events selected
-      const eventPoint = (nextProps.app.selected.length > 0) ? nextProps.app.selected[0] : null;
-
-      if (eventPoint !== null && eventPoint.latitude && eventPoint.longitude) {
-        this.map.setView([eventPoint.latitude, eventPoint.longitude]);
-      }
-    }
-    if (hash(nextProps.app.mapBounds) !== hash(this.props.app.mapBounds) && nextProps.app.mapBounds !== null) {
-      this.map.fitBounds(nextProps.app.mapBounds);
+    // Set appropriate zoom for narrative 
+    if (hash(nextProps.app.mapBounds) !== hash(this.props.app.mapBounds)
+      && nextProps.app.mapBounds !== null) {
+        this.map.fitBounds(nextProps.app.mapBounds);
+    } else {
+      if (hash(nextProps.app.selected) !== hash(this.props.app.selected)) {
+        // Fly to first  of events selected
+        const eventPoint = (nextProps.app.selected.length > 0) ? nextProps.app.selected[0] : null;
+  
+        if (eventPoint !== null && eventPoint.latitude && eventPoint.longitude) {
+          this.map.setView([eventPoint.latitude, eventPoint.longitude]);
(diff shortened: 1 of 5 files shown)


breezykermo:
I'm not sure that this works robustly for me (in Firefox).

Sometimes when selecting a narrative, the map does not appropriately locate the first event, and occasionally precludes relevant events.

The zoom on the timeline also does not keep the selected event in scope unless the selected event is perfectly centred. For example, if I select an event, then drag the timeline so that the selected event is off-center, and then change the intensity of the timeline zoom, no selected events remain in scope. I think the solution here is to programmatically center the first (temporally) selected event programmatically before a zoom intensity is adjusted.

breezykermo:
The timeline also has funny styling for me switching in and out of narrative mdoe. Have you noticed this @francamps? Sometimes the SVG content is not appropriately centred due to the appearing/disappearing `<Toolbar/>`.

francamps:
In fact, the issue may be more of a design one. What are the desired bounds of the map if: i) it has to show all the events in a comfortable way and ii) it also has to center around the the first event in the narrative. At the moment also there seems to be a bit of a race condition between the two, since they are animated and have no callback, which behaves differently in webkit and FF.
I can try to combine i) and ii) and see how it behaves.

With regards to the timeline zoom, I'm not sure that improves the experience. If a user has selected an event, and then drags aways from it being in the center, and then zooms, I think we should assume that that is what she wants to do, and snapping back to the event to center the zoom may be confusing. For example, maybe I have selected an event and want to inspect a different part of the timeline before clicking onto a new event in that other part and snapping back onto the first event would prevent me from doing that.

Ah, I see what you mean about the timeline not being centered. I'll fix that now.

breezykermo:
Played around with this a little further. The loop isn't infinite; but it does sometimes now take a couple of seconds to switch into narrative mode, and/or to move between events. When moving between events in a narrative, the map updates, but then isn't interactive for a couple of seconds...

I'd have to imagine that the computation to work out sensible defaults for the map is sometimes hanging somewhere. It happens for me in both Chrome and Firefox.

breezykermo:
Great; could you rebase your changes on top of the remote develop branch so that I can review before merge?

francamps:
What is the status of this review?

breezykermo:
It currently breaks some of the functionality for me, may be something to do with the most recent changes in other PRs. I'm taking a look

breezykermo:
I'm fairly sure that this works as expected now. Feel free to merge when you're comfortable @francamps

PR #81, opened 2019-01-08
Topic/restyle narrative card

### LINKED ISSUE #45: make Card more materialistic
Cards currently are brought up 'folded', and can be unfolded to display dynamic fields such as source and associated narratives.

It would be nice to make this design flatter, as the folded structure feels a bit weighty (see discussion in #44).

There still needs to be a notion of highlighting a particular card, as it makes map and timeline space more negotiable.

### LINKED ISSUE #70: restyle narrative card
The narrative card needs to be a fixed height to allow the cardstack to easily fall directly beneath it, which doesn't vibe with its current design. Narratives with longer meta text cause it to turn to scroll.

The next/prev arrows for the narrative card should appear at the top of the card (and possibly even could expand to the full screen, to indicate more that we are narrative _mode_). The text padding can be thinner, and prettier when scrolling is necessary.

breezykermo:
Further removes narrative control interactions from `NarrativeCard`. `NarrativeCard` is now a non-interactive display of narrative metadata.

Movement forward and back in the narrative is relegated to left and right `NarrativeAdjust`s, which sit attached above the timeline at either end of the screen. I'm not attached to this visual layout for controls, but I do think that the controls should be taken out of `NarrativeCard` and made prominent in some similar way. Causing left/right arrow presses to trigger `DECREMENT_NARRATIVE_CURRENT` and `INCREMENT_NARRATIVE_CURRENT` actions respectively would be a nice UX improvement.

closes #45 
closes #70

screen shot 2019-01-08 at 18 06 57


francamps:
I really like the style of the narrative card. However, why do we need to take the controls out of the NarrativeCard? It's the more prominent element once you enter narrative mode. Maybe we can simply make them larger within the NarrativeCard? My sense is that there is a bit of conceptual leap to be made between these new controls somewhere else in the screen and the concept of the narrative. By the timeline, they look like they are going to control the timeline itself. 

We may need to rethink this a little, since now that I think of it, now that all the cards of the narrative are showing there is a mismatch: the narrative controls move left and right but the cards go up and down. We should bind them visually and conceptually together a bit more.

As an aside, the icons of the controls are not showing in Chrome:
screen shot 2019-01-09 at 11 07 20 am

breezykermo:
I've pushed a commit to:
* fix the awkward display on Chrome
* add a close button for narrative mode (missing previously)
* rethink placement of narrative adjust controllers.

The reason that I am hesitant to return the narrative adjustment controls to the card is purely out of space constraints. If we put them in the little header that I've added for exiting narrative mode, I worry that they wouldn't be prominent enough.

Very open to alternative placement/styles for narrative adjust. Styling should be easy and intuitive to modify if you want to have a go @francamps, as everything for the prev/next buttons are contained in the `NarrativeAdjust` component.

PR #84, opened 2019-01-16
Topic/categories as tags

### LINKED ISSUE #77: Categories should be available as filterable tags
By default, any partion or annotation of events should become available as a tag to allow filtering by that annotation.

francamps:
This PR implements the feature by which users can filter events by category. 

The categories are displayed in the same panel as tags, in a separate section.

Category and tag filters intersect. This means that if one category C is selected, only events belonging to C will be displayed. If additionally some tags (t1, t2) are then selected, only events belonging to C, tagged t1&t2 will then be displayed.

If no categories are selected, all events are in display.

closes #77

### DIFF
--- a/src/components/Toolbar.jsx
+++ b/src/components/Toolbar.jsx
@@ -81,7 +81,7 @@ class Toolbar extends React.Component {
             categories={this.props.categories}
             tagFilters={this.props.tagFilters}
             categoryFilters={this.props.categoryFilters}
-            filter={this.props.filter}
+            onFilter={this.props.methods.onFilter}
             language={this.props.language}
           />
         </TabPanel>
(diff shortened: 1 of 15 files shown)


src/components/Toolbar.jsx
```diff
@@ -150,15 +151,15 @@ class Toolbar extends React.Component {
...
 
     return (
       <div className="toolbar">
         <div className="toolbar-header"><p>{title}</p></div>
         <div className="toolbar-tabs">
           {/*this.renderToolbarTab(0, 'search')*/}
-          {this.renderToolbarTab(0, 'Narratives')}
-          {(isTags) ? this.renderToolbarTab(1, 'Explore by tag') : ''}
+          {this.renderToolbarTab(0, 'Narratives', 'timeline')}
+          {(isTags) ? this.renderToolbarTab(1, 'Explore by tag', 'style') : ''}
```
breezykermo: Should these strings ('Narratives' and 'Explore by tag') be imported from copy rather than being inline?

src/selectors/index.js
```diff
@@ -54,6 +67,17 @@ function isNoTags(tagFilters) {
   )
 }
 
+/*
+* Returns true if no categories are selected
+*/
+function isNoCategories(categories) {
+  return (
+    categories.length === 0
+    || !process.env.features.CATEGORIES_AS_TAGS
```
breezykermo: Nice, a very concise way to express the feature toggle in one line.

breezykermo:
I've also added a couple of commits that make progress towards #40. Not sure that we can close it completely with this PR, though.

breezykermo:
Also one other minor thing: could narratives stay in the same order that they come in from the API, rather than being alphabetically sorted? Seems an appropriate fix here given its making changes in the Toolbar area.

PR #88, opened 2019-01-17
Source overlay refine

francamps:
This PR takes care of refining the styles of the SourceOverlay component, adding a carrousel for image and video galleries, as well as styling copy for better legibility.

It also includes a variety of small fixes.


breezykermo:
Noting also the more major change that I've made to custom configuring the store. Values are now deep merged into the store on build from 'config.js'. The values specified in 'config.js' `store` attribute take precedence over those in 'src/store/initial.js', so that all custom specification can be done in 'config.js' without having to touch other files.

I'll update the 'example.config.js' file to reflect breaking changes.

PR #89, opened 2019-01-18
Topic/shapes support

### LINKED ISSUE #2: Support for GeoJSON paths
Paths, in [geojson](http://geojson.org/) format, can be used to designate trajectories of interest on the map. In previous iterations, geojson paths have been called `routes`, but the name can be revisited. `routes` have the following properties:

* `routes`' data format are arrays of coordinates, and are assumed to be open. This means that the last coordinate in the array will NOT have an edge connecting it to the first coordinate. They also can have directionality, and will be styled with no fill.
* `routes` should be capable of being a function of time, or as well as being time-independent. 
* `routes` should be capable of being grouped by `category`, and being as well as containing `tags`.
* Like the rest of the application, `routes` will be vector graphics, able of being styled, with properties such as stroke. It is recommended for their fill to be null.
* `routes` could contain additional information to be displayed on click, in the CardStack component, similar to how `events` do.
* `routes` will be fetched together with the rest of the domain.
* Their format will be validated in the reducer together with the rest of the domain. 

screen shot 2018-10-29 at 4 26 57 pm

### LINKED ISSUE #3: Support for displaying communications: multi-location events.
Events of communication can be described as events that happen simultaneously in two or more locations (if there are more than two participants). We suggest to call this `coevents` to refer to this property.

* `coevents` data format should accommodate for multiple sets of coordinates.
* `coevents` are a function of time, similarly to how `events` are. 
* `coevents` should be displayed distinctively on the timeline and on the map
* `coevents` should be capable of being grouped by category, and being as well as containing tags.
* Like the rest of the application, `coevents` will be rendered vector graphics, able of being styled, with properties such as stroke and fil
* `coevents` could contain additional information to be displayed on click, in the CardStack component, similar to how events do.
* `coevents` will be fetched together with the rest of the domain.
* Their format will be validated in the reducer together with the rest of the domain.

screen shot 2018-10-29 at 4 28 21 pm

breezykermo:
Adds the ability to draw custom SVG 'shapes' on the map. Shapes are specified at a separate endpoint (`SHAPES_EXT`) and are only enabled if this is provided and the `USE_SHAPES` flag is true.

I've also refactored the structure of presentational components, to make the repository cleaner in general. Functional components specific to some part of the application such as Map, Timeline, or Narrative are now contained in subfolder within the 'presentational' folder, and are not prefixed with the name of their parent as previously. Several components that were classes have been refactored to functions, and the interface for components in 'Map' has been better standardised. 

`projectPoint` is now specified as a class function in 'Map' and passed directly as a prop to children that need it, rather than the previous solution that had the code repeated four times in children, being composed from the same three props that were passed (which were not used elsewhere, and have now been removed).

Also moves the dimensions of the Timeline to the redux store, so that they can be configured through 'config.js' like other custom values. Minor style adjustment to the forward and back arrows to ensure that no text is occluded.

Added support for custom mapbox backdrops. Defaults to open street map if no Mapbox token given.

Fixed a bug where the timeline labels would not correspond to the appropriate colours.

closes #2


breezykermo:
The commit I just pushed should have fixed.

Issue #36, opened 2018-12-05
Better define 'highlight' logic


breezykermo:
There is a difference between selecting an event and highlighting it, but highlighting is not well defined in Timemap yet: actions are dispatched to the store, but they don't do very much. @francamps, could you comment further on the difference between the two, and their use cases?

PR #96, opened 2019-02-01
Fix/user testing bugs

### LINKED ISSUE #75: Rethink switching between narratives inside narrative mode
The original interaction, clicking the trajectory of another narrative to switch to it, seems to be broken (although it would be easy enough to fix).

That interaction, however, may need some rethinking, given that it is no longer given that mutual implication in a narrative will be represented by lines between events.

@francamps, any ideas?

francamps:
A series of bug fixes and styling refinements are included in this PR. 
- Sources' overlay spacing issues have been fixed.
- Tags are now called filters, and are set in different tabs as Categories. They both have new icons.
- Movement types are optionally custom styled from config
- Timeline has option for hard limits from config.
- Event dots in map display color slices based on aggregate category events they contain
- Cards display precision level for both time and location.

closes #75

### DIFF
--- a/example.config.js
+++ b/example.config.js
@@ -20,19 +20,26 @@ module.exports = {
   },
   store: {
     app: {
-      mapAnchor: [31.356397, 34.784818],
-      filters: {
-        // timerange: [
-        //   new Date(2015, 7, 9),
-        //   new Date(2015, 10, 6, 23)
-        // ]
-      }
+      map: {
+        anchor: [31.356397, 34.784818]
+      },
+      timeline: {
+        range: [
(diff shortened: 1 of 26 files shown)


breezykermo:
A couple of things:
* I haven't set up Travis yet, sorry. Could  you run a `yarn lint` manually and fix any errors?
* Do you think it makes sense to switch all `tags` in the code to `filters`? I do feel that filters are a more appropriate monikor for what we have been referring to as tags.

breezykermo:
Potential close of #36 as well.

PR #98, opened 2019-02-07
Add key and optional cover

### LINKED ISSUE #93: Add configurable legend, shows on application open
A legend based on that in https://plataforma-ayotzinapa.org.

It should include a button to toggle it on and off at any point during application use.

Color-to-category descriptions need to be dynamically templated. Much of the other text can be copy text.

breezykermo:
* Adds a toggleable info-popup that is shown by default to explain the available interactions in Timemap.
* Add an optional cover, which is designed to give context to a Timemap instance. There is a default cover that shows how to set one up, but the idea is that each instance would likely create a custom cover (just a React component) to detail the instance. Is shown by default if `USE_COVER` feature flag is enabled.

closes #93

### DIFF
--- a/src/components/Dashboard.jsx
+++ b/src/components/Dashboard.jsx
@@ -139,7 +139,9 @@ class Dashboard extends React.Component {
         <InfoPopUp
           ui={ui}
           app={app}
-          toggle={() => actions.toggleInfoPopup()}
+          methods={{
+            onClose: actions.toggleInfoPopup
+          }}
         />
         <Notification
           isNotification={app.flags.isNotification}
--- a/src/components/Dashboard.jsx
+++ b/src/components/Dashboard.jsx
@@ -14,6 +14,7 @@ import InfoPopUp from './InfoPopup.jsx'
 import Timeline from './Timeline.jsx'
 import Notification from './Notification.jsx'
 import StaticPage from './StaticPage'
+import DefaultCover from './presentational/covers/Default'
 
 import { parseDate } from '../js/utilities'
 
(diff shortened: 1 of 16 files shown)


src/scss/infopopup.scss
```diff
@@ -24,87 +24,108 @@
...
-      &.cocula-municipal-police { background-color: #095959; }
-      &.ambulance { background-color: #ffffff; }
-      &.other { background-color: #D3CE2A; }
-      &.drivers { background-color: #822519; }
-      &.communications { background-color: #a6a6a6; }
-      &.GIEI { background-color: #ffffff; }
-      &.PGR { background-color: #000000; }
-    }
-  }
+  //     &.victims { background-color: #C90500; }
```
francamps: Let's just remove this stuff

francamps: Like all of it. Users can define their own styling if they want something more sophisticated in the legend.

src/components/Dashboard.jsx
```diff
@@ -155,6 +159,13 @@ class Dashboard extends React.Component {
             }
           />
         ) : null}
+        {process.env.features.USE_COVER && (
+          <StaticPage showing={app.flags.isCover}>
+            {/* enable USE_COVER in config.js features, and customise your header */}
+            {/* pass 'actions.toggleCover' as a prop to your custom header */}
+            <DefaultCover showAppHandler={actions.toggleCover} />
+          </StaticPage>
```
francamps: Great stuff.

PR #111, opened 2019-05-16
Topic/template cover

samludford:
Adds a template cover which can be configured via the config file. It includes an updated example.config.js which shows how to configure the cover.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,9 +3,6 @@ build/
 node_modules/
 config.js
 dev.config.js
-# ignore all covers but the default
-src/components/presentational/covers/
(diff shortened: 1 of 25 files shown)


breezykermo:
Looks good in general. I think presentational/covers/TemplateCover.js was not pushed due to it being in the .gitignore. Could you explicitly add it, remove that exclusionary logic in the .gitignore, and push again?

breezykermo:
Looking better! I think you also want to add a default value for `app.cover` in 'src/store/initial.js'. I'm currently getting an error in the console, as I don't have any custom `cover` values defined in my config.

PR #115, opened 2019-05-22
Topic/aesthetic review

samludford:
- yellow hover state on location markers
- dotted ring on selected location yellow
- narrative boxes in toolbar yellow hover state


src/components/presentational/Map/SelectedEvents.jsx
```diff
@@ -13,7 +13,7 @@ class MapSelectedEvents extends React.Component {
       >
         <path
           className='leaflet-interactive'
-          stroke={styles ? styles.stroke : '#ffffff'}
+          stroke={styles ? styles.stroke : '#ffd800'}
```
breezykermo: Let's make this an imported variable that comes from a 'js/global.js'. Have just noticed that the 'js' folder should probably be renamed to 'common' as well.  The `FA_RED` variable in 'src/store/initial.js' can come from this as well

breezykermo: A couple more hardcoded colours that you might nab while at it:

src/components/Toolbar/Layout.js
64:              <button style={{ backgroundColor: '#000' }} onClick={() => { this.goToNarrative(narr) }}>

src/components/presentational/Map/SelectedEvents.jsx
16:          stroke={styles ? styles.stroke : '#ffffff'}

src/components/presentational/Map/Events.jsx
18:      stroke: '#ffffff',

src/components/presentational/Timeline/Markers.js
10:        stroke={styles ? styles.stroke : '#ffffff'}

src/store/initial.js
3:const FA_RED = '#eb443e';

PR #125, opened 2019-08-19
Make d3 an npm-managed dependency

### LINKED ISSUE #91: Make d3 an npm-managed dependency
The library d3 is currently being loaded in `index.html` from CDN at Cloudflare. Just like the rest of dependencies in the application, it should be a npm dependency.

delucis:
Closes #91

This pull request moves from using a `<script>` tag to load `d3` on the global scope to importing `d3` to the application bundle from the package published to npm.

I ran the `timemap` development server alongside the `datasheet-server` example documented [here](https://engineering.forensic-architecture.org/timemap-and-datsheet-server/) and everything looks to be working the same, but someone with more familiarity with the project should probably double check that no behaviour has changed.


breezykermo:
LGTM! Thanks so much, I'll merge :).

PR #132, opened 2020-04-04
Markdown file container respects line breaks

frasercrichton:
The line breaks in the linked markdown files are not currently respected. This might be one way to fix that.


breezykermo:
Nice catch, thanks.

PR #136, opened 2020-04-20
Topic/time only events

breezykermo:
- switches layered events from displaying numbers to being a set opacity, so that the overall event space reads like a heatmap.
- significantly fixes the dynamic size of the timeline according to store values.
- adds a feature that views non-located events as a 'graph', where projects are sets of events that are shifted on the y-axis according to how many projects are contemporaneous.


breezykermo:
Note: need to rename things here so that it makes more sense in general. Maybe start documenting features as well.

breezykermo:
Merging this for now, will submit a PR shortly to clean up things...

PR #142, opened 2020-06-25
display source titles

lostground:
display the source title instead of its id in the card


breezykermo:
Thanks! This is the expected behavior, but I think the display should default to `source.id` if `source.title` doesn’t exist, as the second is optional and can be blank. (Quite often we just collapse the two, but you’re right in that I think I originally introduced it so that you could make the display different from the ID if wished.)

Once you make that simple change I’ll merge. :)

lostground:
It looks like this change was made in yesterday's updates.

breezykermo:
Ah yes, apologies. I implemented a splurt of updates yesterday and completely lost track of the fact that I had made this adjustment as well.

PR #145, opened 2020-07-14
Remove unnecessary if statement

sol-austin:
I've removed an unnecessary if statement from index.js because the if statement checks if the variable if present, however it is inside a forEach loop, so it must be present anyway.

I've also commented out minTime and maxTime because they caused the timeline to show 1970-present when using the ilovaisk spreadsheet.


breezykermo:
The `minTime` and `maxTime` lines are correct here. They are used in other instances of timemap when there are maxBounds set on the timeline. 

I'm not sure why they would need to be commented out to get the Ilovaisk sheet working, but I imagine it may be because of something awry in the config.js.

If you revert 53cee7c I can merge these changes. Thanks!

PR #146, opened 2020-07-14
Add narrative clickability

sol-austin:
I have added an action and reducer to allow a direct change in current narrative step by idx, rather than incrementing or decrementing.

I had to add mapDispatchToProps to some files to allow action calls from the component.

The map doesn't seem to always consistently work. Sometimes when clicking, the correct narrative card is chosen, however sometimes it is not.


breezykermo:
Hiya- sorry for the delay here.

First thing- I've realised that the issue you may have been running into around `appState.timeline.rangeLimits` is that there is no default `rangeLimits` set on the redux state, and so features that make use of it require explicit setting in the config. I've just pushed a commit to the develop branch that fixes this- so if you rebase onto develop, you'll hopefully not run into those errors any longer.

On the actual feature here, this is a great start! I'll review your additions more in depth inline, but one general point to begin with is that, when a user clicks a card in the narrative cardstack on the right, the appropriate event should become highlighted in both the timeline and the map (in the same way as it does when you click on a different point in the map). Additionally, all of the clickable elements (cards, and inactive events in both the timeline and map) should have a `cursor: pointer` on hover in the CSS, so that it's obvious that they can be clicked.

src/actions/index.js
```diff
@@ -260,6 +260,14 @@ export function decrementNarrativeCurrent () {
...
 }
 
+export const SELECT_NARRATIVE_EVENT = 'SELECT_NARRATIVE_EVENT'
+export function selectNarrativeEvent (idx) {
+  return {
+    type: SELECT_NARRATIVE_EVENT,
+    idx
+  }
+}
+
```
breezykermo, line 270: I'd suggest a slightly different architecture here that is more closely tuned to how narrative navigation currently works.

Notice what happens currently when one of the arrows is pressed in narrative mode: an `UPDATE_SELECTED` action is dispatched, followed by an `INCREMENT_NARRATIVE_CURRENT` or `DECREMENT_NARRATIVE_CURRENT`, followed by an `UPDATE_TIMERANGE` action. 

I would suggest refactoring the `INCREMENT_NARRATIVE_CURRENT`/`DECREMENT_NARRATIVE_CURRENT` into a more general action, maybe `UPDATE_NARRATIVE_IDX`, and then dispatching this action (along with `UPDATE_SELECTED` and `UPDATE_TIMERANGE` if appropriate) whenever a card or event in the narrative is pressed.  The `UPDATE_NARRATIVE_IDX` action can also then replace the current dispatches of `INCREMENT_NARRATIVE_CURRENT` and `DECREMENT_NARRATIVE_CURRENT` (i.e., those will become `updateNarrativeIdx(currentIdx + 1)` and `updateNarrativeIdx(currentIdx-1)`, with appropriate checks for corner cases, respectively).

These actions could all probably be managed by an action creator which is called from the components, which in turn dispatches the appropriate actions as per above.

src/components/CardStack.jsx
```diff
@@ -185,4 +194,10 @@ function mapStateToProps (state) {
...
 }
 
-export default connect(mapStateToProps)(CardStack)
+function mapDispatchToProps (dispatch) {
+  return {
+    actions: bindActionCreators(actions, dispatch)
+  }
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(CardStack)
```
breezykermo, line 203: You'll notice that [several methods are passed as props to CardStack from Layout](https://github.com/forensic-architecture/timemap/blob/develop/src/components/Layout.js#L278-L286). I would prefer that, instead of connecting this component directly to the redux store as you've done, we stick with the current setup here and pass the appropriate function down as a prop.

src/components/Map.jsx
```diff
@@ -202,6 +205,11 @@ class Map extends React.Component {
     return [null, null]
   }
 
+  onSelectNarrativeEvent (idx) {
+    console.log(idx)
```
breezykermo, line 209: remove `console.log`.

src/components/Map.jsx
```diff
@@ -298,4 +306,10 @@ function mapStateToProps (state) {
...
 }
 
-export default connect(mapStateToProps)(Map)
+function mapDispatchToProps (dispatch) {
+  return {
+    actions: bindActionCreators(actions, dispatch)
+  }
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(Map)
```
breezykermo, line 315: Likewise here- the methods should be passed as props ([see here](https://github.com/forensic-architecture/timemap/blob/develop/src/components/Layout.js#L262-L269)), rather than the component being connected directly to the store. The reason is that this makes the redux methods that these components use much more explicit.

sol-austin:
I added the CSS cursor: pointer so it is clear the cards are now clickable. 

I also refactored the code so access to actions isn't directly needed from any child components and most logic is completed in Layout.js.

A drawback of my solution is the logic in Map/Events.jsx. It always uses index 0 when selecting an event on the map, meaning if multiple events in a narrative are in one location, the first one is selected. This means that the narrative state is always consistent as it only one event is selected at a time. 

I removed the INCREMENT and DECREMENT narrative counters from the actions and reducers and refactored the Layout.js call to use the idx directly.

src/components/Layout.js
```diff
@@ -178,16 +179,26 @@ class Dashboard extends React.Component {
...
+      const idx = current + 1
+      this.selectNarrativeEvent(idx)
+    }
+    if (amt === -1) {
+      const idx = current - 1
+      this.selectNarrativeEvent(idx)
+    }
+  }
+
+  selectNarrativeEvent (idx) {
```
breezykermo, line 193: I'd suggest renaming this function `selectNarrativeStep`, to make it more clear what is happening.

src/components/Layout.js
```diff
@@ -178,16 +179,26 @@ class Dashboard extends React.Component {
 
   moveInNarrative (amt) {
```
breezykermo, line 180: I'd probably suggest getting rid of this function altogether, and instead just calling the increment directly in the handlers. That is, replace all instances of `this.moveInNarrative(vl)` with `this.selectNarrativeEvent(this.props.app.narrativeState.current + vl)`. It will reduce the code bloat.

src/components/Layout.js
```diff
@@ -264,7 +275,8 @@ class Dashboard extends React.Component {
           methods={{
             onSelect: ev => this.handleSelect(ev, 1),
             onSelectNarrative: this.setNarrative,
-            getCategoryColor: this.getCategoryColor
+            getCategoryColor: this.getCategoryColor,
+            onSelectNarrativeEvent: this.selectNarrativeEvent
```
breezykermo, line 279: The `onSelectNarrativeEvent` prop actually doesn't need to exist here. Instead, the functionality should just be passed down as a conditional in the `onSelect` handler. This will simplify the logic inside the component as well.

For example:
```js
methods = {
  onSelect: app.narrative ? ev => this.handleSelect(ev, 1) : this.selectNarrativeEvent
}
```

src/components/Layout.js
```diff
@@ -283,6 +295,7 @@ class Dashboard extends React.Component {
           onToggleCardstack={() => actions.updateSelected([])}
           getNarrativeLinks={event => this.getNarrativeLinks(event)}
           getCategoryColor={this.getCategoryColor}
+          onSelectNarrativeEvent={this.selectNarrativeEvent}
```
breezykermo, line 298: The same here. There shouldn't be a new prop, it should instead be passed via a conditional to the `onSelect` prop.

src/components/Map.jsx
```diff
@@ -214,7 +214,7 @@ class Map extends React.Component {
         selected={this.props.app.selected}
         narrative={this.props.app.narrative}
         onSelect={this.props.methods.onSelect}
-        onSelectNarrative={this.props.methods.onSelectNarrative}
+        onSelectNarrative={this.props.methods.onSelectNarrativeEvent}
```
breezykermo, line 217: And here as well: call it conditionally from the `onSelect` handler.

src/components/presentational/Map/Events.jsx
```diff
@@ -135,7 +144,7 @@ function MapEvents ({
       <g
         className={`location-event ${narrative ? 'no-hover' : ''}`}
         transform={`translate(${x}, ${y})`}
-        onClick={(!narrative) ? () => onSelect(location.events) : null}
+        onClick={(!narrative) ? () => onSelect(location.events) : () => { onSelectNarrative(narrativeIdx); onSelect(location.events) }}
```
breezykermo, line 147: This logic will be simplified by passing to the `onSelect` props as suggested above.

src/components/presentational/Map/Events.jsx
```diff
@@ -110,13 +111,21 @@ function MapEvents ({
 
     // in narrative mode, only render events in narrative
     // TODO: move this to a selector
+    // Gets around if block scope
+    var narrativeIdx = false
```
breezykermo, line 115: In general, prefer `let` to `var`- see [here](https://www.geeksforgeeks.org/difference-between-var-and-let-in-javascript/) on why (`var` behaves in a charismatic way which introduces more complexity for programmers).

breezykermo: Is there a reason that this default value is `false`? Would it be better as `null`? Or `-1`?

src/reducers/app.js
```diff
@@ -9,8 +9,7 @@ import {
   UPDATE_TIMERANGE,
   UPDATE_DIMENSIONS,
   UPDATE_NARRATIVE,
-  INCREMENT_NARRATIVE_CURRENT,
-  DECREMENT_NARRATIVE_CURRENT,
+  SELECT_NARRATIVE_IDX,
```
breezykermo, line 12: I would change this to `UPDATE_NARRATIVE_IDX`, or event `UPDATE_NARRATIVE_STEP_IDX`. In general when you are adding actions to a Redux application, you want to take a look at the other actions, and try to mirror the use of verbs in the action names for consistency.

breezykermo:
I've just tested this using the example config. There seems to be a small bug:
* when I am in a narrative, and I select an event _from the timeline_, the selected event on the map updates, but +the active card in the card stack does not update_. 

Clicking events from other parts of the app seem to work as expected (i.e. clicking on the card appropriately updates both timeline and map, and clicking on an event in the map appropriately updates both the timeline and card stack).

sol-austin:
I've refactored the code so Layout.js is more logic heavy than the components. Also, the bug should be fixed!

src/actions/index.js
```diff
@@ -246,17 +246,11 @@ export function updateNarrative (narrative) {
   }
 }
 
-export const INCREMENT_NARRATIVE_CURRENT = 'INCREMENT_NARRATIVE_CURRENT'
-export function incrementNarrativeCurrent () {
+export const UPDATE_NARRATIVE_STEP_IDX = 'UPDATE_NARRATIVE_STEP_IDX'
+export function selectNarrativeIdx (idx) {
```
breezykermo, line 250: This can also become `updateNarrativeStepIdx` to keep consistent.

src/components/Layout.js
```diff
@@ -176,18 +176,37 @@ class Dashboard extends React.Component {
     })
   }
 
-  moveInNarrative (amt) {
-    const { current } = this.props.app.narrativeState
+  selectNarrativeStep (idx) {
+    // Try to find idx if event passed rather than number
+    if (typeof idx !== 'number') {
+      console.log('run')
+      console.log(idx)
```
breezykermo, line 183: remove

src/components/Layout.js
```diff
@@ -176,18 +176,37 @@ class Dashboard extends React.Component {
...
+      if (this.props.app.narrative) {
+        const { steps } = this.props.app.narrative
+        // choose the first event at a given location
+        const locationEventId = e.id
+        const narrativeIdxObj = steps.find(s => s.id === locationEventId)
+        let narrativeIdx = steps.indexOf(narrativeIdxObj)
+
+        if (narrativeIdx > -1) {
+          idx = narrativeIdx
+          console.log(idx)
```
breezykermo, line 195: remove

src/components/Layout.js
```diff
@@ -176,18 +176,37 @@ class Dashboard extends React.Component {
...
+        }
+      }
+    }
+
+    console.log(idx)
```
breezykermo, line 200: remove

src/components/Layout.js
```diff
@@ -176,18 +176,37 @@ class Dashboard extends React.Component {
     })
   }
 
-  moveInNarrative (amt) {
-    const { current } = this.props.app.narrativeState
+  selectNarrativeStep (idx) {
+    // Try to find idx if event passed rather than number
+    if (typeof idx !== 'number') {
```
breezykermo, line 181: this is a nice way of getting around the fact that `onSelect` is called with an event as argument, rather than an index.

Another way to have done this would have been to create a `getIndexInNarrative` function (something like this may already exist in utils or selectors, in fact), and to call this inside the lines that pass the function as a prop to child components. This would keep `selectNarrativeStep` cleaner in terms of its argument type- `idx` would always be an integer. (In many programming languages, functions will require that you explicitly type function arguments.)

I don't mind this way, though!

src/reducers/app.js
```diff
@@ -98,19 +97,8 @@ function updateNarrative (appState, action) {
...
-    narrativeState: {
-      current: appState.narrativeState.current
-    }
-  }
-}
-
-function decrementNarrativeCurrent (appState, action) {
-  appState.narrativeState.current -= 1
+function updateNarrativeStepIdx (appState, action) {
+  appState.narrativeState.current = action.idx
```
breezykermo, line 101: because reducers are [pure functions](https://redux.js.org/introduction/three-principles), you normally want to avoid modifying the state that is passed in. The simple solution here is just to inline `action.idx` directly in line 106 below, where the new state is returned.

PR #151, opened 2020-08-04
Add search bar

sol-austin:
I've added a couple of files for the Search and SearchRow components as well as a CSS stylesheet. The events are searchable by category, date, location and description and the search term is highlighted. Only 5 words on each side of the search term are shown to minimise space used.

The main problem at the moment is a slight lag when typing in the search bar. Because the search value is sent to the redux store on change and then retrieved, and passed down through components, it seems if users type quickly, it takes a moment for the words to actually appear in the input field.

### DIFF
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -243,7 +243,7 @@ class Dashboard extends React.Component {
   }
 
   updateSearchQuery (e) {
-    let queryString = e.target.value;
+    let queryString = e.target.value
     this.props.actions.updateSearchQuery(queryString)
   }
   render () {
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -21,6 +21,7 @@ import TemplateCover from './TemplateCover'
 import colors from '../common/global'
 import { binarySearch, insetSourceFrom, findDescriptionInFilterTree } from '../common/utilities'
 import { isMobile } from 'react-device-detect'
+import Search from './Search.jsx'
 
 class Dashboard extends React.Component {
   constructor (props) {
@@ -240,6 +241,11 @@ class Dashboard extends React.Component {
       }
     }
   }
+
+  updateSearchQuery (e) {
+    let queryString = e.target.value;
+    this.props.actions.updateSearchQuery(queryString)
+  }
   render () {
     const { actions, app, domain, ui, features } = this.props
 
@@ -325,6 +331,12 @@ class Dashboard extends React.Component {
           isNotification={app.flags.isNotification}
           notifications={domain.notifications}
           onToggle={actions.markNotificationsRead}
+        />
+        <Search
+          narrative={app.narrative}
+          queryString={app.searchQuery}
+          onQueryUpdate={this.updateSearchQuery}
+          events={domain.events}
         />
         {app.source ? (
           <MediaOverlay
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -71,6 +71,7 @@ class Dashboard extends React.Component {
   }
 
   handleSelect (selected, axis) {
+    console.log(selected)
     const matchedEvents = []
     const TIMELINE_AXIS = 0
     if (axis === TIMELINE_AXIS) {
@@ -337,6 +338,7 @@ class Dashboard extends React.Component {
           queryString={app.searchQuery}
           onQueryUpdate={this.updateSearchQuery}
           events={domain.events}
+          onSearchRowClick={this.handleSelect}
         />
         {app.source ? (
           <MediaOverlay
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -71,7 +71,6 @@ class Dashboard extends React.Component {
   }
 
   handleSelect (selected, axis) {
-    console.log(selected)
     const matchedEvents = []
     const TIMELINE_AXIS = 0
     if (axis === TIMELINE_AXIS) {
@@ -243,10 +242,6 @@ class Dashboard extends React.Component {
     }
   }
 
-  updateSearchQuery (e) {
-    let queryString = e.target.value
-    this.props.actions.updateSearchQuery(queryString)
-  }
   render () {
     const { actions, app, domain, ui, features } = this.props
 
@@ -336,7 +331,6 @@ class Dashboard extends React.Component {
         <Search
           narrative={app.narrative}
           queryString={app.searchQuery}
-          onQueryUpdate={this.updateSearchQuery}
           events={domain.events}
           onSearchRowClick={this.handleSelect}
         />
--- a/src/components/Map.jsx
+++ b/src/components/Map.jsx
@@ -92,6 +92,7 @@ class Map extends React.Component {
     firstLayer.addTo(map)
 
     map.keyboard.disable()
+    map.zoomControl.remove()
 
     map.on('move zoomend viewreset moveend', () => this.alignLayers())
     map.on('zoomstart', () => { if (this.svgRef.current !== null) this.svgRef.current.classList.add('hide') })
--- a/src/components/Search.jsx
+++ b/src/components/Search.jsx
@@ -1,0 +1,67 @@
+import React from 'react'
+
+import '../scss/search.scss'
+
+import SearchRow from './SearchRow.jsx'
+
+class Search extends React.Component {
+    constructor(props) {
+        super(props)
+        this.state = {
+            isFolded : true,
+            searchResults: [],
+            queryString: ''
+        }
+        this.onButtonClick = this.onButtonClick.bind(this)
+        this.updateSearchQueryResults = this.updateSearchQueryResults.bind(this)
+    }
+
+    componentDidUpdate (prevProps, prevState) {
+        if (prevProps.queryString !== this.props.queryString) {
+            this.updateSearchQueryResults (this.props.queryString)
+        }
+    }
+
+    onButtonClick () {
+        this.setState(prevState => {
+            return { isFolded : !prevState.isFolded }
+        })
+    }
+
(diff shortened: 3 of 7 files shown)


breezykermo:
Great thanks, I'll take a look at this shortly.

One way that you can improve the responsiveness of an input with search like this is by throttling the response, only sending it when the user 'stops typing'. See [this thread](https://stackoverflow.com/questions/4220126/run-javascript-function-when-user-finishes-typing-instead-of-on-key-up) for a brief description. (We probably wouldn't want the 'stops typing' threshold to be 5s, but something more like 1-2s.)

src/components/SearchRow.jsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  function getShortDescription (text, searchQuery) {
+    var regexp = new RegExp(`(([^ ]* ){0,6}[a-zA-Z]*${searchQuery.toLowerCase()}[a-zA-Z]*( [^ ]*){0,5})`, 'gm')
+    let parts = text.toLowerCase().match(regexp)
+    for (var x = 0; x < (parts ? parts.length : 0); x++) {
+      parts[x] = '...' + parts[x]
+    }
+    const firstLine = [text.match('(([^ ]* ){0,10})', 'm')[0]]
+    return parts || firstLine
```
breezykermo, line 18: Really nice stuff! Perhaps some of this could be abstracted into pure functions in the 'utilities.js'. This way, the functionality is ready to be reused in other components if necessary. It also means you can give some more descriptive names to the regex (in the function name), which just makes the code a little more readable. :)

breezykermo (approved):
Hey there- this is really good stuff, I'm very sorry for the extended delay in my review!

I've made a few inline comments about rearchitecting the `<Search/>` component slightly, but otherwise this is a really great addition that I'm more or less ready to merge. The only minor improvement I'd suggest, if at all possible, is to come up with a throttling strategy (see an earlier comment I made on this PR) in order to prevent the experience where typing quickly freezes up the app for a small while.

src/components/Search.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+  onButtonClick () {
+    this.setState(prevState => {
+      return { isFolded: !prevState.isFolded }
+    })
+  }
+
+  updateSearchQueryResults (queryString) {
+    let searchResults
+    if (queryString === '') {
```
ebefarooqui, line 33: Empty strings are already falsey in JS so you can just check `if (!queryString)`

src/components/Search.jsx
```diff
@@ -0,0 +1,67 @@
...
+      searchResults = []
+    } else {
+      searchResults = this.props.events.filter(event =>
+        event.description.toLowerCase().includes(queryString.toLowerCase()) || event.location.toLowerCase().includes(queryString.toLowerCase()) || event.category.toLowerCase().includes(queryString.toLowerCase()) || event.date.includes(queryString)
```
ebefarooqui, line 37: These conditionals could be modularized and less verbose. Maybe destructure these from the event object and then iterate over them and set a boolean accordingly. Then check the boolean as opposed to running the same logic through each variable in one line.

src/components/Layout.js
```diff
@@ -70,6 +72,7 @@ class Dashboard extends React.Component {
   }
 
   handleSelect (selected, axis) {
+    console.log(selected)
```
ebefarooqui, line 75: Is this needed?

src/scss/search.scss
```diff
@@ -0,0 +1,110 @@
+#search-bar-icon-container {
```
ebefarooqui, line 1: Now that we are moving into a design system, should we spec out designs more thoroughly for this @breezykermo ? Good work here btw!

src/components/SearchRow.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+
+const SearchRow = ({ query, eventObj, onSearchRowClick }) => {
+  const { description, location, date } = eventObj
+  function getHighlightedText (text, highlight) {
+    // Split text on highlight term, include term itself into parts, ignore case
+    const parts = text.split(new RegExp(`(${highlight})`, 'gi'))
+    return <span>{ parts.map(part => part.toLowerCase() === highlight.toLowerCase() ? <span style={{ backgroundColor: 'yellow', color: 'black' }}>{part}</span> : part) }</span>
```
ebefarooqui, line 8: This is great, but it would be good to have more consistent indentation here for legibility

src/components/Search.jsx
```diff
@@ -0,0 +1,67 @@
...
+      )
+    }
+    this.setState({
+      searchResults: searchResults
+    })
+  }
```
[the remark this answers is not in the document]
ebefarooqui: Agreed, generally it's good to avoid having to use `setState` unless absolutely necessary. The render / state change cycle is a better way to handle that logic.

Also, as a general note for all of our development, the most recent React documentation warns against using some of the older lifecycle methods like `componentDidUpdate`. I think it'd be best for us to avoid using them moving forward, but I'm definitely open to discussing the matter.

https://reactjs.org/docs/react-component.html

src/components/Map.jsx
```diff
@@ -92,6 +92,7 @@ class Map extends React.Component {
     firstLayer.addTo(map)
 
     map.keyboard.disable()
+    map.zoomControl.remove()
```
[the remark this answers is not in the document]
sol-austin: It removes the zoom control buttons in the top left which the search icon is replacing.

src/components/Search.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+  componentDidUpdate (prevProps, prevState) {
+    if (prevProps.queryString !== this.props.queryString) {
+      this.updateSearchQueryResults(this.props.queryString)
+    }
+  }
+
+  onButtonClick () {
+    this.setState(prevState => {
+      return { isFolded: !prevState.isFolded }
```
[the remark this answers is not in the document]
sol-austin: `isFolded` controls whether the search bar panel is open or not whereas `queryString` and `searchResults` are linked to the search query and results directly. I could probably remove either `queryString` or `searchResults` from the state as they are directly linked with each other. Maybe just store `queryString` in the state and have the Search component find the results based on the query when it is rendered?

src/components/Layout.js
```diff
@@ -226,6 +229,11 @@ class Dashboard extends React.Component {
       }
     }
   }
+
+  updateSearchQuery (e) {
+    let queryString = e.target.value
+    this.props.actions.updateSearchQuery(queryString)
```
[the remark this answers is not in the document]
sol-austin: Yep sure. That might even mean that all Search state involving results can be handled within the search component itself. As you say the `handleSelect` once a result is clicked needs to be in the parent component but getting the results could be contained within the component.

src/components/Search.jsx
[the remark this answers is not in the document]
sol-austin: Yeah, sure. I'll try to refactor to minimise state use and avoid lifecycle methods. I thought `componentDidUpdate` is okay and `componentWillUpdate` is no longer used, or are they both being phased out?

src/components/Layout.js
sol-austin: I always seem to leave one `console.log`!

sol-austin:
I've refactored so only the query string is stored in state. Also removed some repetition and moved more login into the search component.

src/scss/search.scss
breezykermo: Yes, I think we can look internally at how to redesign this and think about moving it into the design system being developer. I don't think it should be a blocker on merging this PR, though! I think it adds to timemap as written here in the meantime. :)

breezykermo:
@sol-austin great, thank you! Are you still making changes, or is it ready to be reviewed again?

sol-austin:
> @sol-austin great, thank you! Are you still making changes, or is it ready to be reviewed again?

I might have a look at sorting the typing delay as you said, but if that isn't a blocker to this I'd prefer to do it in a separate PR

src/components/Search.jsx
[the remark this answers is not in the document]
breezykermo: I think you're right @sol-austin: it's just `componentWillUpdate` that is now legacy.

breezykermo:
I've just taken another look at this. @sol-austin another essential feature that I think we would need before merge is to filter search results _only_ to those that have a matching string. I'm noticing that currently the cards are not filtered, the matching text is just highlighted. This will be a problem for any non-trivial timemap, as there will be many more event cards than can display in the search results.

I think may be as simple as changing a `.map` to `.filter` in Search.jsx. Do you have time to make the change? If not, I can create a new PR based off this branch and add the changes. :)

sol-austin:
How odd. I don't think it was doing that in a previous commit. I'll sort it!

sol-austin:
Only had to remove a pair of brackets and switch a 'map' to a 'some' method. It's sorted now on my Timemap instance.

breezykermo:
Great! Would you be able to rebase the branch onto the latest `develop`, and merge any conflicts? Then I can merge. :)

sol-austin:
Already pulled and rebased I think?

breezykermo:
I believe it needs to be done again sorry. I’m getting an error indicating that the branch cannot be automatically rebased due to conflicts.

sol-austin:
Well that was quite messy. I did a rebase from develop from the main repo. Hopefully that's resolved the conflicts.

breezykermo:
Screen Shot 2020-09-21 at 1 54 44 PM

Unfortunately I'm still getting this! I think you need to run one last `git merge --no-ff add-search-bar` as the github instructions above suggest.

The search filters as expected now, thanks!

Issue #10, opened 2018-11-04
Create example instance, add screenshot in README


breezykermo:
Create a live example instance that uses an instance [datasheet-server](https://github.com/forensic-architecture/datasheet-server) to present information from a public sheet.

Would also be nifty to put a screenshot in the README to provide a visual description of the software.

breezykermo:
This needs to include a document regarding how we set up 'EXPORT' tabs in sheets, to demonstrate end-to-end how to use both technologies in tandem

PR #159, opened 2020-09-11
Feature/refactor filters and narratives to associations

ebefarooqui:
Associated Issue: [#10](https://github.com/forensic-architecture/tasks/issues/10)
Associated `datasheet-server` [PR](https://github.com/forensic-architecture/datasheet-server/pull/54) 

This PR refactors the following:

**1. Collapse usage of filters and narratives into one set called associations**

We've decided to rebrand 'filters' and 'narratives' as a single data type called 'associations'.

An association can be imagined as a thread that collects together multiple events. An association can either be viewed as a filter, or as a narrative.

The way that associations are encoded in a sheet are with the following attributes:

```
{
  "id": "The name of the association",
  "desc": "A description of what the association collects",
  "mode": "The default way in which the association shows, 'filter' or 'narrative'",
  "filter_path": ["Grandparent", "parent"]
}
```

The filter_path is an ordered list that dictates how the association is showed when it is being viewed as a filter.

**2. Deprecate usage of narratives and filters for events and collapse into associations**

**3. Deprecate usage of redundant environment variables in config sheet => replaced with values interpolated from associations object**


breezykermo:
The CI we have here doesn't check much, but it does check linting!

You can run `npm run lint` to apply it to the relevant files.

ebefarooqui:
Fixed linting issues

ebefarooqui:
@breezykermo @zacoppotamus I modified the US Protests sheet and tested these changes with those modifications. Everything works accordingly, but maybe one of you could double check.

zacoppotamus:
This seems to work on my end. Only small error I get is in the front-end with an individual event:
`events 145: Invalid date. It's been dropped, as otherwise timemap won't work as expected.`

image

ebefarooqui:
> This seems to work on my end. Only small error I get is in the front-end with an individual event:
> `events 145: Invalid date. It's been dropped, as otherwise timemap won't work as expected.`
> 
> image

This is a validation check that is run on the data to see if the row's date is correctly formatted. It's a data-related alert and not particular to the changes made in this PR. Sanity check on this? @breezykermo

breezykermo (changes requested):
One major thing that seems to be broken: when clicking on the source dropdown in the card, nothing opens, and the selected events gets deselected.

src/common/utilities.js
```diff
@@ -199,19 +199,6 @@ export function binarySearch (ar, el, compareFn) {
...
-export function findDescriptionInFilterTree (key, node) {
-  if (node.key === key) return node.description
-  if (isFilterLeaf(node)) return false
-  const descs = Object.keys(node.children)
-    .map(c => findDescriptionInFilterTree(key, node.children[c]))
-    .filter(v => !!v)
-  if (descs.length !== 1) return false
-  return descs[0]
-}
-
```
breezykermo, line 214: Nice- I'm very glad to get rid of this logic!

src/components/Toolbar/FilterListPanel.js
```diff
@@ -3,57 +3,72 @@ import Checkbox from '../presentational/Checkbox'
...
+function childrenToToggle (filter, activeFilters, parentOn) {
+  const [key, children] = filter
+  const isOn = activeFilters.includes(key)
+  if (children === {}) {
+    return [key]
   }
-  const childKeys = Object.values(node.children)
-    .flatMap(n => childrenToToggle(n, activeFilters, isOn))
+  const childKeys = Object.entries(children)
+    .flatMap(filter => childrenToToggle(filter, activeFilters, isOn))
```
breezykermo, line 13: Could you explain why it is that we still have the recursive logic here? My understanding from our discussions is that, because associations are now 'flat', there shouldn't be any need for recursion. Instead, we should just be able to work out the layout structure once up front using `filter_paths`, and then just render it accordingly.

This should also mean that the logic of working out when associations should be active/inactive whenever something is clicked is not recursive, but simply a matter of filtering the (flat) list of associations for those that contain the relevant values in `filter_paths`.

src/components/Toolbar/FilterListPanel.js
```diff
@@ -3,57 +3,72 @@ import Checkbox from '../presentational/Checkbox'
...
+      if (!(path in aggregated)) {
+        currentDepth[path] = {}
+      }
+      currentDepth = currentDepth[path]
+    })
+  })
+
+  return aggregated
+}
+
```
breezykermo, line 38: I tried adding another level in the 'EXPORT_ASSOCIATIONS' tab in timemap_data, to make sure I understood how it was working... and I'm not sure it worked. 

This is how I updated the tab:
Screen Shot 2020-09-17 at 11 01 59 AM

I would have thought that this would cause another level in the filters tab, but nothing seems to have changed:
Screen Shot 2020-09-17 at 11 02 32 AM

The data is coming into the domain correctly, from datasheet-server, but it is not affecting the layout of the filter panel:
Screen Shot 2020-09-17 at 11 03 24 AM

src/reducers/validate/validators.js
```diff
@@ -162,20 +125,16 @@ export function validateDomain (domain, features) {
...
+  // Duplicated associations
+  if (duplicateAssociations.length > 0) {
     sanitizedDomain.notifications.push({
-      message: `Filters are required to be unique. Ignoring duplicates for now.`,
-      items: duplicateFilters,
+      message: `Associations are required to be unique. Ignoring duplicates for now.`,
+      items: duplicateAssociations,
       type: 'error'
     })
   }
```
breezykermo, line 136: I want to make sure I understand correctly what's going on here as well. In the old version of filters, we did not allow duplicate names for filters, _even if they were hierarchised differently in the tree layout_. This was essentially due to the recursive logic that was implemented to find filter names when toggling them, as we didn't have any notion of filters being distinguishable by their level in the filter tree: they were only distinguishable by their ID.

If I understand correctly, the requirement that we now have is just that the `id` of an association is unique. This `id`, however, doesn't explicitly affect its position in the filter tree layout, as this is now wholly determined by the `filter_paths` attribute.

Associations _can_, in contrast, have identical `filter_paths` arrays, an equivalence which seems to essentially collapse two associations into the same filter from a UI perspective. My original thought was that the `id` (or another attribute we could add called `name`) would always be appended to the end of the `filter_paths` array, so that each association appears as a unique way to filter events: but I actually might prefer this way that you have set it up, as it gives us more flexibility around how associations show up as filters.

After thinking/writing through this logic, perhaps it makes sense to change `filter_paths` to just `grouping`. It does determine how the filters are displayed, but it is also a representation of the 'grouping' in which each individual association falls. These grouping buckets can extend to arbitrary depth, which gives us flexibility in how the filters are laid out.

These association groupings would also then be the technical translation of 'fields of investigation' in the 3D timeline feature @zacoppotamus. The people would become associations, and the fields in which they participate (Mexico, Saudi Arabia, etc) would then be their groupings. Although this means we would need a slightly different representation in the sheet, as people can be part of multiple groupings... perhaps I'll open another issue to think this through.

src/components/Toolbar/FilterListPanel.js
ebefarooqui: Fixed

ebefarooqui:
> One major thing that seems to be broken: when clicking on the source dropdown in the card, nothing opens, and the selected events gets deselected.

Fixed. The source of this bug is related to the changes pushed in the search bar PR

src/reducers/validate/validators.js
ebefarooqui: Reading through this and seeing that the request is to make a semantic change from `filter_paths` to `groupings`. Is that your desire here (beyond the discussion) or can we move forward without the name change? @breezykermo

src/components/Toolbar/FilterListPanel.js
ebefarooqui: I tried an iterative solution but it made more sense to do it recursively because the set of filter paths overlap when laid out structurally. In order to account for children being set appropriately when their parents are selected, we need to have a noted relationship between the parent node of the tree and the child. For example, if we have `Weapons/Close-Range/Tear-Gas` and `Weapons/Long-Range`, the tree would appear as
`[
    { Weapons: { Close Range: { Tear Gas: {} }, Long Range: {} } }
  ]`
I'd be keen to know whether or not there is a non-tree / recursive way of spinning up that representation. I could be missing something here.

src/reducers/validate/validators.js
breezykermo: No worries, can go forward without. Was just riffing

src/selectors/index.js
```diff
@@ -1,24 +1,24 @@
...
 import { isTimeRangedIn } from './helpers'
+import { FILTER_MODE, NARRATIVE_MODE } from '../common/constants'
 
 // Input selectors
 export const getEvents = state => state.domain.events
 export const getCategories = state => state.domain.categories
-export const getNarratives = state => state.domain.narratives
-export const getActiveNarrative = state => state.app.narrative
-export const getActiveStep = state => state.app.narrativeState.current
+export const getNarratives = state => state.domain.associations.filter(item => item.mode === NARRATIVE_MODE)
```
breezykermo, line 9: This is something to talk about. In the way that these selectors currently work, an association can only have one mode: either `NARRATIVE_MODE` or `FILTER_MODE`. 

I think it should be the case that an association can display, in fact, in multiple modes, and that its `mode` attribute merely designates the dominant mode in which it displays... or perhaps a more elegant solution is to make `modes` an array, so that you could display an association as both a narrative _and_ a filter. I think this will become particular relevant when we have the third way to display an association, as `TIMELINE`.

But just something to think about for the next category PR, it shouldn't block this!

breezykermo (approved):
This is such a welcome PR! I think transforming filters and narratives to associations is a much more elegant approach, evidenced by how much crusty code is able to be excised with the new abstraction. :) looks very good to me, just one comment to think about for the next PR on categories.

src/selectors/index.js
ebefarooqui: Taking note of this. I agree with you on making modes an array so that there can be multiple existing modes for a given association. It'll allow for more flexibility moving forward with our current notion of associations and whether or not we add more allowable `types` for an association.

PR #161, opened 2020-09-23
Feature/multi select events

ebefarooqui:
This PR does the following:

**1) Allow for the user to multi select events**
    Configured so that the user can shift and select events

**2) Selecting an event doesn't modify set of events displayed and timeline remains the same**
   Previously, selecting an event on the map would shift the timeline accordingly to center the event on the timeline.
   This meant that the selection of events displayed on the map would change as well. This PR removes the aforementioned logic.


breezykermo (approved):
Very nice, short and sweet

PR #162, opened 2020-10-06
Feature/add supercluster visualization

ebefarooqui:
This is a preliminary PR to get feedback for the supercluster visualization. Currently, the styles are still in question (opacity, size of cluster, cluster radius) and we will be reviewing different styles soon to gauge what is best for the user.

**See comment below for more updated information**


breezykermo:
Love this feature. Do you want general comments regarding what's missing and bugs, or just inline comments on the code reviewing what you've done so far?

src/components/Map.jsx
```diff
@@ -94,14 +116,64 @@ class Map extends React.Component {
...
+        clusters: this.index.getClusters(bbox, zoom)
+      })
+    }
+  }
+
+  loadClusterData (locations) {
+    if (locations && locations.length > 0 && this.index) {
+      const convertedLocations = locations.reduce((acc, loc) => {
+        const { longitude, latitude } = loc
+        const validCoordinates = !!latitude && !!longitude
```
breezykermo, line 152: you could use `isLatitude` and `isLongitude` from common/utilities here to check more extensively and make it a bit more readable.

src/components/Map.jsx
```diff
@@ -94,14 +116,64 @@ class Map extends React.Component {
...
+            properties: {
+              cluster: false,
+              id: loc.label
+            },
+            geometry: {
+              type: 'Point',
+              coordinates: [longitude, latitude]
+            }
+          }
+          acc.push(feature)
```
breezykermo, line 165: nice use of `reduce`!

src/components/Map.jsx
```diff
@@ -39,6 +48,9 @@ class Map extends React.Component {
   }
 
   componentWillReceiveProps (nextProps) {
+    if (hash(nextProps.domain.locations) !== hash(this.props.domain.locations)) {
```
breezykermo, line 51: We could put this in a function in common/utilities now, e.g. `!isIdentical(nextProps.domain.locations, this.props.domain.locations)`, which just uses the hash under the hood. I've thought about this being an externality we could prune out at some point, so it might be sensible to add that layer of abstraction so it's easy to prune in the future.

src/components/Map.jsx
```diff
@@ -94,14 +116,64 @@ class Map extends React.Component {
...
+              type: 'Point',
+              coordinates: [longitude, latitude]
+            }
+          }
+          acc.push(feature)
+        }
+        return acc
+      }, [])
+      this.index.load(convertedLocations)
+      this.setState({ indexLoaded: true })
```
breezykermo, line 170: what appears on the map before the clusters are loaded? I see in `update` that `state.clusters` will remain an empty list. If it defaults to the old representation (events without clusters), would it perhaps be preferable to show a loading icon?

src/components/presentational/Map/Clusters.jsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  function renderClusterBySize (cluster) {
+    const { point_count: pointCount, cluster_id: clusterId } = cluster.properties
+    const { coordinates } = cluster.geometry
+    const [longitude, latitude] = coordinates
+
+    const totalPoints = calculateTotalPoints()
+
+    const styles = ({
```
breezykermo, line 30: no `()` needed

src/components/presentational/Map/Clusters.jsx
```diff
@@ -0,0 +1,101 @@
...
+    const [longitude, latitude] = coordinates
+    if (!latitude || !longitude) return null
+    const { x, y } = projectPoint([latitude, longitude])
+
+    const customStyles = styleCluster ? styleCluster(cluster) : null
+    const extraRender = () => (
+      <React.Fragment>
+        {customStyles[1]}
+      </React.Fragment>
+    )
```
breezykermo, line 78: This `extraRender` logic is a good candidate for making into a HOC now, I think, given that it's used both here and in Events.jsx.

breezykermo (changes requested):
Really excellent work/feature! I've made a couple of structural suggestions around code quality. I haven't been able to test the feature yet in a browser, due to having the wrong config I think (have sent you a message privately).

ebefarooqui:
This PR should be fully ready to review. I've attached the appropriate timemap config file [here](https://github.com/forensic-architecture/configs/blob/master/timemap/us_protests_cluster_config.js).

Note that the categories are currently broken on this branch, but you shouldn't need to test with them anyways. The styling for the clusters (opacity, size, cluster radius) are the ones decided on in the chat, but any and all suggestions welcome.

In order to run locally with updated data, I created a `test_export_events` tab in the datasheet, so in order to see the data locally, make sure that the return statement in `src/lib.js` in `ds-s` looks like the following:

`return {`
   `[`test_export_events`]: BP.deeprows,`
   `[`${prf('categories')}export_categories`]: [BP.groups, BP.rows],`
   `[`${prf('associations')}export_associations`]: BP.deeprows,`
   `[`${prf('sources')}export_sources`]: BP.deepids,`
   `[`${prf('sites')}export_sites`]: BP.rows`
  `}`

src/common/utilities.js
```diff
@@ -173,7 +173,7 @@ export function calcClusterOpacity (pointCount, totalPoints) {
 }
 
 export function calcClusterSize (pointCount, totalPoints) {
-  return Math.min(50, 10 + (pointCount / totalPoints) * 250)
+  return Math.min(50, 10 + (pointCount / totalPoints) * 150)
```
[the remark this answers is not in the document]
ebefarooqui: They are essentially multiplication factors. Will leave a comment above

src/components/Map.jsx
```diff
@@ -11,9 +12,13 @@ import 'leaflet'
 import Sites from './presentational/Map/Sites.jsx'
 import Shapes from './presentational/Map/Shapes.jsx'
 import Events from './presentational/Map/Events.jsx'
+import Clusters from './presentational/Map/Clusters.jsx'
 import SelectedEvents from './presentational/Map/SelectedEvents.jsx'
 import Narratives from './presentational/Map/Narratives'
 import DefsMarkers from './presentational/Map/DefsMarkers.jsx'
+import DefsClusters from './presentational/Map/DefsClusters.jsx'
```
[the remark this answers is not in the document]
ebefarooqui: I was just testing out the radial gradients and so `Defs` is a remnant of that. I've left it for now, in case we ever specify in the config that we'd like radial gradients. Will also move it out into the `Clusters` component.

src/components/Map.jsx
```diff
@@ -23,11 +28,15 @@ class Map extends React.Component {
   constructor () {
     super()
     this.projectPoint = this.projectPoint.bind(this)
+    this.onClusterSelect = this.onClusterSelect.bind(this)
     this.svgRef = React.createRef()
     this.map = null
+    this.index = null
```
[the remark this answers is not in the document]
ebefarooqui: gotcha

src/components/Map.jsx
```diff
@@ -39,6 +48,9 @@ class Map extends React.Component {
   }
 
   componentWillReceiveProps (nextProps) {
+    if (hash(nextProps.domain.locations) !== hash(this.props.domain.locations)) {
+      this.loadClusterData(nextProps.domain.locations)
```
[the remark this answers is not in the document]
ebefarooqui: I do want to move a lot of the logic out, but it is tightly coupled with the Map component. I could scope it outside of the Map class, but then it would need access to instances of the map and index that are declared within that component. Will look into a better refactor soon, but tabling for now.

src/components/Map.jsx
ebefarooqui: Definitely, will need to add some sort of interim state so that the events are all loaded and then displayed.

PR #165, opened 2020-10-21
Feature/supercluster corners

ebefarooqui:
Fixes the following:

1) Bug: initial map render not displaying initial clusters
2) Bug: zooming into location devoid of clusters throws a black screen with an error
3) Bug: Scrolling on timeline until all clusters disappear on timeline isnt reflected in map state


breezykermo (commented):
Nice!

PR #166, opened 2020-10-22
Bugfix/make supercluster independently configurable

ebefarooqui:
Moving supercluster config outside of the map. This is important because we might want to use a separate set of attributes for the clusters as we do the map.

### DIFF
--- a/src/components/Map.jsx
+++ b/src/components/Map.jsx
@@ -79,7 +79,7 @@ class Map extends React.Component {
     /**
      * Creates a Leaflet map and a tilelayer for the map background
      */
-    const { map: mapConfig } = this.props.app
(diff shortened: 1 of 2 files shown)


src/components/Map.jsx
```diff
@@ -90,9 +90,9 @@ class Map extends React.Component {
 
     // Initialize supercluster index
     this.superclusterIndex = new Supercluster({
-      radius: mapConfig.clusterRadius,
-      maxZoom: mapConfig.maxZoom,
-      minZoom: mapConfig.minZoom
+      radius: clusterConfig.radius,
+      maxZoom: clusterConfig.maxZoom,
+      minZoom: clusterConfig.minZoom
```
breezykermo, line 95: you could just do `new Supercluster(clusterConfig)` here, no?

PR #167, opened 2020-10-22
Feature/ux fixes

breezykermo:
@ebefarooqui changes from that screen share here.

### DIFF
--- a/src/components/Map.jsx
+++ b/src/components/Map.jsx
@@ -89,11 +89,7 @@ class Map extends React.Component {
         .setMaxBounds(mapConfig.maxBounds)
 
     // Initialize supercluster index
-    this.superclusterIndex = new Supercluster({
-      radius: clusterConfig.radius,
-      maxZoom: clusterConfig.maxZoom,
-      minZoom: clusterConfig.minZoom
-    })
+    this.superclusterIndex = new Supercluster(clusterConfig)
 
     let firstLayer
 
@@ -116,6 +112,7 @@ class Map extends React.Component {
     map.zoomControl.remove()
 
     map.on('moveend', () => {
+      // console.log(map.getZoom())
       this.updateClusters()
       this.alignLayers()
     })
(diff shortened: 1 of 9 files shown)


src/components/Map.jsx
```diff
@@ -287,12 +282,12 @@ class Map extends React.Component {
...
   renderEvents () {
     const individualClusters = this.state.clusters.filter(cl => !cl.properties.cluster)
-    const filteredLocations = mapClustersToLocations(individualClusters, this.props.domain.locations)
+    // const filteredLocations = mapClustersToLocations(individualClusters, this.props.domain.locations)
     return (
       <Events
         svg={this.svgRef.current}
         events={this.props.domain.events}
-        locations={filteredLocations}
+        locations={this.props.domain.locations}
```
ebefarooqui, line 290: So this is us rendering everything in locations? Good to note for later

src/components/presentational/Map/Clusters.jsx
```diff
@@ -12,97 +12,108 @@ const DefsClusters = () => (
...
+      point_count: int,
+      point_count_abbreviated: int
+    },
+    type: "Feature"
+  }
+  */
+  const { cluster_id: clusterId } = cluster.properties
+  const { coordinates } = cluster.geometry
+  const [longitude, latitude] = coordinates
+  if (!latitude || !longitude) return null
```
ebefarooqui, line 33: can use the utility functions here for longitude and latitude

ebefarooqui (approved):

src/components/Map.jsx
breezykermo: Yeah I was just experimenting with this, but it adds the dots beneath the clusters as Eyal mentioned. It also means clusters doesn't give us the performance gains that we talked about: so I think I'll actually revert it back. Perhaps we can make it a feature toggle/store value?

ebefarooqui: That makes sense. This will affect the way that the selected are rendered on the map, no?

breezykermo: No, because the `<Markers/>` are rendered from a different component. So these will still be drawn even when the events themselves aren't rendered.

PR #170, opened 2020-10-27
Fix/recursive corner case

breezykermo:
This PR fixes two corner cases introduced by the coloring set algorithm that can cause the state of active filters and color to get way out of whack. (There is undefined and hard-to-interpret behavior whenever an incorrect action is dispatched for toggling filters, or for updating the coloring sets, as these two aspects of the state are tightly coupled.)

1. The `getFilterSiblings` function was always returning an empty array for meta filters (implicit filters that exist as parents in `filter_paths`, but not as associations in the sheet). This is because the function was only checking the `id`s of assocations, and not recursing/reducing up the `filter_path` array to find implicit filters. `getFilterSiblings` now delegates to `getMetaFilterSiblings` when the filter key passed is not an association, and returns the appropriate siblings. This was causing havoc in terms of appropriately toggling off grandparents when parent sets had been toggled off after toggling on a grandparent.

2. Similarly, the control flow that toggles off a paren when its last child is toggled off was not seeking further up the family tree, and appropriately toggling off grandparents as well if necessary. This also caused coloring havoc- now fixed.

### DIFF
--- a/src/common/utilities.js
+++ b/src/common/utilities.js
@@ -117,9 +117,29 @@ export function getImmediateFilterParent (associations, filter) {
 }
 
 /**
+ * Grab a meta filter's siblings, by way of the the `filter_path` hierarcy.
+*/
+export function getMetaFilterSiblings (allFilters, filterParent, filterKey) {
+  const idxParent = allFilters.map(f => {
+    return f.filter_paths.reduceRight((acc, path, idx) => {
+      if (path === filterParent) return f.filter_paths[idx + 1]
+      return acc
+    }, null)
+  })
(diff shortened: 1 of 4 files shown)


breezykermo:
@ebefarooqui feel free to merge once you've reviewed. @zacoppotamus and I agreed that we can dramatically simplify the logic in components to make this more maintainable, most probably by 'serializing' the `filter_path` hierarchy to a tree-like structure when the domain is first loaded and keeping this in the Redux store. Then we can use this as the canonical representation and recurse down it to find filters: rather than having to always deal with the two variants of filters, association-filters and meta-filters, at every point in the control flow.

ebefarooqui (approved):

PR #182, opened 2020-11-12
Added generators for card layouts. 

zacoppotamus:
I figured adding a sibling generator file be the least invasive way. The generators are optionally defined in the timemap config, and have also been tested with the Yemen config (and the appropriate `datasheet-server` branch).  I'm using the optional chaining JS operator to ascertain the existence of the card layout setting in the redux-connected props — this feature is included in `@babel/plugin-preset-env@7.8.0`, which we have already. The linting fails because of this, but once we merge the `feature/build-system` branch, this won't be a problem as it uses [Prettier](https://prettier.io/) for linting.


breezykermo:
What's the plan to migrate this hardcoded `us2020` generator out of the codebase? Could it just go as a feature toggle in 'config.js' for the time being, instead of being inside the timemap source code?

breezykermo (changes requested):
Is there any reason that we can't move the `us2020` generator into the us2020 config? I would rather keep aspects of particular instances from the timemap src code.

breezykermo:
I also think it would make more sense to move the `basic` generator inside the Redux store 'intial.js' default. It's currently inside the 'components' folder, however it's not a React component.

src/components/CardStack.jsx
```diff
@@ -1,179 +1,180 @@
...
-      const thisRef = React.createRef();
-      this.refs[idx] = thisRef;
+      const thisRef = React.createRef()
+      this.refs[idx] = thisRef
 
       return (
         <Card
           ref={thisRef}
-          content={generateLayouts[this.props.cardUI?.layout || `basic`]({
+          content={this.props.cardUI.layout({
```
zacoppotamus, line 71: Here is the relevant change @breezykermo

zacoppotamus: Apologies for the crowded diff — it's due to having reverted to our current linting stack

zacoppotamus:
I believe this should address the requested changes @breezykermo. I moved the basic card layout inside the `common` folder, am calling it in `initial.js`, and am overriding changes through the config.

src/common/card.js
```diff
@@ -0,0 +1,31 @@
...
+    [
+      {
+        kind: 'text',
+        title: 'Summary',
+        value: event.description || ``,
+        scaleFont: 1.1
+      }
+    ]
+  ]
+}
```
breezykermo, line 27: nice! I like this cleanliness a lot more. So the current implementation requires a generator comparable to this one (specifying custom `value`s if they exist on the event object) to exist in the config?

breezykermo (approved):

src/common/card.js
breezykermo: If so, could you make an equivalent PR to the configs repo?

zacoppotamus: There's a pending PR with these changes in the configs repo :)

PR #184, opened 2020-11-20
Snowpack-based build system

zacoppotamus:
I'd like to open this PR and start a conversation around what we want our build process (and by extension developer experience) to look like for timemap.

My top priorities in this branch were to move to a faster build system (Snowpack) that optimizes by not re-bundling unchanged files in `node_modules`, and hinges on modern browsers' in-built support of ES modules. I chose Snowpack as it seems to be quite a mature project that's getting picked up by [Svelte](https://svelte.dev/) as well.

There were quite a few changes that needed to be made:
- Rename files to `.jsx`. For some reason, this was necessary for Snowpack (can't recall the exact reason unfortunately). This doesn't affect our Webpack bundle process though
- Bump year-old versions of `joi` and `object-hash` to their isomorphic (work the same in a Node vs JS environment) equivalents and update they way they're called in our application
- Move our configs inside `src/` rather than have them ingested into our code, as was previously done by Webpack
- Enforce linting on file save. This is done through Prettier, as I noticed our current linter don't support some later ES6 syntax, and also I couldn't set up our previous linter in VSCode to format directly on save. For this reason, I've been responsible for big diffs in commits I've made in the past 😢. To me, this format-on-save process has become an integral part of the coding experience and it's especially useful when working with anonymous and nested functions (which show up quite a bit in JSX).
- Formatted everything using Prettier

For another project, I've been using `react-scripts@4.x.x` which offers a very pleasant refresh experience (still Webpack under the hood). If this is our fallback option, it takes care of a lot of things under the hood, but still requires of us that we either deprecate our current config ingestion or `npm eject` the `create-react-app` and change the Webpack configuration ourselves. 

Things that remain to be done or discussed:
- [ ] remove the web_modules folder (this is the intermediary for package consumption that Snowpack generates)
- [ ] merge with the latest changes in the `develop` branch
- [ ] configure `react-refresh` correctly so that it enables HMR that retains our Redux store across saves
- [ ] decide what the best way of ingesting our configs into the codebase are, and how to ensure it works the same both in the dev cycle and also in the build cycle. In the future we want to hydrate app-specific configuration through our store anyway (and `datasheet-server`) so this point will become obsolete, but in the meantime we need to make a call
- [ ] decide what we want our build process to look like: should it be handled by Parcel, Snowpack, or Webpack? To be clear, this refers to the process that generates the final bundle that is consumed.


breezykermo:
Thanks for this! I'm not opposed at all to moving to Snowpack in principle, and I agree that it would be ideal to have a dev server with hot reloading, especially as timemap ramps up into more complexity. I believe that this should be possible within Webpack as well; we just haven't taken a look at the dev process for some time.

Linting-on-save seems to me mostly a developer-specific thing, as it depends on what code editor one is using. The current linting is [standard](https://standardjs.com/), which [has a corresponding plugin in VScode](https://marketplace.visualstudio.com/items?itemName=chenxsan.vscode-standardjs) with lint on save. This should work, as I don't believe that there is any specific customization for linting: it just runs the default `standard` command with baked-in rules. 

I'm not entirely sure how I'm supposed to run/test this, and what to give feedback on. I tried `rm -rf node_modules` followed by `npm install`; but after this `npm run dev` fails to compile with:
```
ERROR in ./src/components/App.jsx
Module not found: Error: Can't resolve './Layout' in '/home/lachlan/code/fa/timemap/src/components'
 @ ./src/components/App.jsx 3:0-30 7:44-50
 @ ./src/index.jsx
```

To the point about `web_modules`: is there a reason that it can't just be added to .gitignore, if it is generated by the Snowpack build/dev process? I've never used Snowpack before, so forgive my ignorance here.

As I said, in principle I'd be happy to move to Snowpack: although we would need a solution that allows ingesting configs from the top-level with an environment variable, in the same way that it is currently done with Webpack. In other words, the commands `CONFIG=custom.config.js npm run dev` and `CONFIG=custom.config.js npm run build` should both still work. Is there something fundamentally uncustomizable about Snowpack that prevents this from being possible?

As to the way that the final build is created; it makes sense to me that it is kept in touch with how the dev process is running. There will of course have to be some different configuration for the production build (minimizing, bundling, etc), but unless there is a good reason for it, I would prefer to keep within one build tool (i.e. just Webpack, or just Snowpack), rather than adding maintainence complexity by using multiple.

breezykermo:
@zacoppotamus closing this due to the alternative fix, we can keep the branch for reference.

PR #187, opened 2020-11-29
Feature/spanish translation

1cgonza:
A first attempt to pull request with Spanish localization. Before moving forward, I am thinking that the `config.js` file will require the `store.app.language` property, or these changes will break an app that skips it. Should we parse/check the config data with some defaults before exporting?

### DIFF
--- a/src/common/utilities.js
+++ b/src/common/utilities.js
@@ -413,12 +413,12 @@ export function makeNiceDate (datetime) {
  * Sets the default locale for d3 to format dates in each available language.
  * @param {Object} d3 - An instance of D3
  */
-export function setD3Locale(d3) {
+export function setD3Locale (d3) {
   const languages = {
     'es-MX': require('./data/es-MX.json')
   }
 
   if (process.env.store.app.language !== 'es-US' && languages[process.env.store.app.language]) {
-    d3.timeFormatDefaultLocale(languages[process.env.store.app.language]);
+    d3.timeFormatDefaultLocale(languages[process.env.store.app.language])
   }
 }
--- a/src/common/utilities.js
+++ b/src/common/utilities.js
@@ -1,6 +1,7 @@
 import moment from 'moment'
 import hash from 'object-hash'
 
+const language = process.env.store.app.language || 'en-US'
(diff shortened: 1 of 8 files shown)


breezykermo:
This looks great, thanks! I think you just have to run `npm run lint:fix` in order to get the build to pass :)

src/common/utilities.js
```diff
@@ -406,5 +406,19 @@ export function makeNiceDate (datetime) {
...
+/**
+ * Sets the default locale for d3 to format dates in each available language.
+ * @param {Object} d3 - An instance of D3
+ */
+export function setD3Locale(d3) {
+  const languages = {
+    'es-MX': require('./data/es-MX.json')
+  }
+
+  if (process.env.store.app.language !== 'es-US' && languages[process.env.store.app.language]) {
```
breezykermo, line 421: It would be good to add code here (and everywhere else in the PR that `process.env.store.app.language` is used) to fall back to `en-US` if an override is not provided in the config.

In other words, users should not have to specify 'en-US' as a language in the config (as I think as it is currently written this code will complain if this is the case, as the `language` attribute will not exist on `process.env.store.app`).

breezykermo: (Correct me if I'm wrong, as I haven't actually run this use case to test :-) )

1cgonza: Yes you are right. I've tested without any language property in config.js and it works now with changes in 2c2b518

ebefarooqui:
Overall looking good! Small nitpick: could you move the line `const language = process.env.store.app.language || 'en-US'` into the utilities file and import it wherever you need it? Just to avoid the duplicate line across files

1cgonza:
> Overall looking good! Small nitpick: could you move the line `const language = process.env.store.app.language || 'en-US'` into the utilities file and import it wherever you need it? Just to avoid the duplicate line across files

@ebefarooqui I like the way you think, take a look at 7faca16 and let me know if that looks good to you.

breezykermo:
LGTM. @ebefarooqui feel free to merge once you're comfortable with this. Thanks @1cgonza!

breezykermo (approved):

PR #188, opened 2020-12-08
New build system with ejected Create React App

zacoppotamus:
This PR moves our build system to an ejected instance of Create React App. Ejecting was necessary in order to specify an external configuration file, which then gets parsed into our app through the `webpack.DefinePlugin` and exposes the JSON object under `process.env`. This allows us to do `CONFIG=../path-to-config.js npm run react-scripts:watch`.

Among other things, this PR features git-commit-hooks for linting (switching over to Prettier), and Hot Module Reloading.


zacoppotamus:
@ebefarooqui @breezykermo I've checked in the necessary files under `config/` — this should be ready for review.

package.json
```diff
@@ -5,59 +5,188 @@
...
   "private": true,
   "scripts": {
-    "dev": "webpack-dev-server --content-base static --mode development",
+    "react-scripts:start": "node scripts/start.js",
+    "react-scripts:build": "NODE_ENV=production node scripts/build.js",
+    "react-scripts:eject": "node scripts/eject.js",
     "dev:wsl": "npm run dev -- --host 0.0.0.0",
-    "build": "NODE_ENV=production webpack --mode production",
     "test": "ava --verbose",
     "test-watch": "ava --watch",
```
breezykermo, line 13: These commands fail for me. I think we worked out that they should be simple enough to clean up, no?

package.json
```diff
@@ -5,59 +5,188 @@
   "homepage": "",
   "private": true,
   "scripts": {
-    "dev": "webpack-dev-server --content-base static --mode development",
+    "react-scripts:start": "node scripts/start.js",
```
breezykermo, line 8: Could we keep the commands the same? I.e. `npm run dev` here, and `npm run build` below.

package.json
```diff
@@ -5,59 +5,188 @@
...
+    "react-scripts:build": "NODE_ENV=production node scripts/build.js",
+    "react-scripts:eject": "node scripts/eject.js",
```
breezykermo, line 10: If we've already run this, do we need to keep it in package.json?

config/env.js
```diff
@@ -0,0 +1,126 @@
...
+
+// Make sure that including paths.js after env.js will read .env variables.
+delete require.cache[require.resolve("./paths")];
+
+/** env variables from config.js */
+const CONFIG = process.env.CONFIG || "config.js";
+const envConfig = require("../" + CONFIG);
+
+const userConfig = {};
+const userFeatures = {};
```
breezykermo, line 15: I'm fairly sure that this logic relating to `userFeatures` can be axed. The features used to come in separately from the rest of the store, in their own standalone object that was saturated separately from config, which is why this logic was necessarily (as the features needed to be available in various parts of the codebase). Then I think I realised that this duplication could be removed by shifting the `features` object to the store.

config/env.js
```diff
@@ -0,0 +1,126 @@
...
+// https://github.com/motdotla/dotenv-expand
+dotenvFiles.forEach((dotenvFile) => {
+  if (fs.existsSync(dotenvFile)) {
+    require("dotenv-expand")(
+      require("dotenv").config({
+        path: dotenvFile,
+      })
+    );
+  }
+});
```
breezykermo, line 55: Is it perhaps worth refactoring the files here in the 'config' folder to clearly separate which code is coming from ejected CRA, and which has been inserted? Perhaps we could have a `cra_setup` or something similar at the root? Not sure if this makes sense, as I'm not fully familiar with what this bundling runtime requires/makes possible, but just a thought.

config/jest/cssTransform.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+// This is a custom Jest transformer turning style imports into empty objects.
+// http://facebook.github.io/jest/docs/en/webpack.html
```
breezykermo, line 4: Is this jest different from the testing framework jest? If not, is there some duplication between this and 'ava' in the `devDependencies`? Happy to cut out ava if it makes sense.

package.json
```diff
@@ -5,59 +5,188 @@
...
+      "web.jsx",
+      "jsx",
+      "node"
+    ],
+    "watchPlugins": [
+      "jest-watch-typeahead/filename",
+      "jest-watch-typeahead/testname"
+    ],
+    "resetMocks": true
+  },
```
breezykermo, line 183: Perhaps we can remove all this if we're not using jest?

public/index.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>TimeMap - Forensic Architecture</title>
+    <link
+      rel="stylesheet"
+      href="https://api.mapbox.com/mapbox.js/v3.1.1/mapbox.css"
+    />
```
breezykermo, line 9: Hmm, this might actually be an unnecessary stylesheet import now that we default to OSM. Perhaps we can make a note to remove it and import it dynamically when Mapbox is toggled on?

scripts/test.js
```diff
@@ -0,0 +1,53 @@
...
+  argv.indexOf('--watchAll') === -1 &&
+  argv.indexOf('--watchAll=false') === -1
+) {
+  // https://github.com/facebook/create-react-app/issues/5210
+  const hasSourceControl = isInGitRepository() || isInMercurialRepository();
+  argv.push(hasSourceControl ? '--watch' : '--watchAll');
+}
+
+
+jest.run(argv);
```
breezykermo, line 53: Can this file go as well, as per excising jest? Or perhaps it's more responsible to port the ava test to jest, given it's already set up to work with CRA webpack infra.

src/common/constants.js
```diff
@@ -1,5 +1,5 @@
 export const ASSOCIATION_MODES = {
-  CATEGORY: 'CATEGORY',
-  NARRATIVE: 'NARRATIVE',
-  FILTER: 'FILTER'
-}
+  CATEGORY: "CATEGORY",
+  NARRATIVE: "NARRATIVE",
+  FILTER: "FILTER",
+};
```
breezykermo, line 5: I'm assuming all of the changes to the actual content here are a matter of linting. If it's not too tricky, maybe we can separate these linting changes from the build system PR? I'm easy either way, though.

breezykermo (approved):
Works really well! A couple of cleanliness notes mostly, but it's a huge improvement to have hot reloading. I haven't tested the build process yet: once @ebefarooqui confirms that we can deploy via the devops scripts, then I'm very happy to merge this.

config/jest/cssTransform.js
zacoppotamus: Yup, I'll remove all Jest-related functionality for the time being.

src/common/constants.js
zacoppotamus: Correct. Other than a couple of new files that were added inside `src/`, the other files remain the same. A way to clean this up would be to remove the git commit hook that lints `src/` and `test/` and re-lint using the old linter. I can try doing this and separate the linting process to a different branch before we merge this PR.

scripts/test.js
zacoppotamus: I ended up including both suites (and updated the README to reflect this). I'm currently using Jest for snapshot testing, whereas the Ava test suites tests the Node process that instantiates the app. Moving forwards, I'd like for us to deprecate the Ava test suite as Jest seems more fully-featured and integrated within the CRA webpack infrastructure. I haven't ported the Ava test suite over as I feel this is outside the scope of this PR. However, both test suites are passing and should hopefully pass our Travis tests.

package.json
zacoppotamus: See below — I've decided to keep both test suites for now and later on deprecate Ava.

config/env.js
zacoppotamus: There's only ~10 lines in this file that relate to ingesting our configurations, so I'm not sure it would make much sense to separate the two. I'll write a comment that makes it clear.

package.json
zacoppotamus: These are passing now. I've made the Jest suite the default for now.

breezykermo:
@zacoppotamus is this ready to be reviewed/merged from your perspective? Have forgotten the status, sorry.

zacoppotamus:
Yes, just waiting for @ebefarooqui input on CI/CD contingencies.

ebefarooqui:
@zacoppotamus Will try this out locally as well and then test against the new CD workflow. Not anticipating too many changes for that integration, although I will probably have to make some in the ansible scripts in the devops repo. I think to a certain extent, we can merge and then accommodate said changes in the CD workflow.

Starting a review now

ebefarooqui (approved):
Running this locally @zacoppotamus and seeing a number of warnings show up. Really minor so in the next PR to develop, I'll introduce some cleanup. Everything else is in the clear, good work!

ebefarooqui:
@zacoppotamus @breezykermo Tangentially related: we should force linting and testing to pass first before allowing merges with any branch.

zacoppotamus:
> @zacoppotamus @breezykermo Tangentially related: we should force linting and testing to pass first before allowing merges with any branch.

I'll make a new PR for removing all the warnings (this currently causes the CI to fail).

PR #194, opened 2021-02-05
Getting rid of duplicate association logic, removing check and error …

ebefarooqui:
Removes check for duplicate associations. Subsequently removes error generation as well

### DIFF
--- a/src/reducers/validate/validators.js
+++ b/src/reducers/validate/validators.js
@@ -24,26 +24,6 @@ function isValidDate(d) {
   return d instanceof Date && !isNaN(d);
 }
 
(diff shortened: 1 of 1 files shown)


breezykermo:
Sorry, I think the issue I created in tasks was unclear. I don't think we should remove this check and associated error message-- we should just modify the example sheet so that it doesn't have duplicate associations by default, so that the error doesn't show up when a new user first starts up a timemap instance.

breezykermo (changes requested):
See above.

ebefarooqui:
Nixing this as it's not relevant

ebefarooqui:
Nixing this as per above

PR #225, opened 2022-03-21
Updates

breezykermo:
- Twitter/Telegram in-card support
- Remove @forensic-architecture/design-system as an NPM dependency, with more generic layouts for cards.
- A range of other general usability fixes.


breezykermo:
@ebefarooqui would you be able to test this with the default timemap/datasheet, and make sure it's reasonable? There are probably a few things that shouldn't be merged to the upstream, so if you can comment I'll roll them back.

src/components/controls/atoms/Media.js
```diff
@@ -0,0 +1,94 @@
...
+  }, []);
+
+  const type = typeForPath(src);
+  const formattedTitle =
+    title && title.length > TITLE_LENGTH
+      ? `${title.slice(0, TITLE_LENGTH + 1)}...`
+      : title;
+
+  switch (type) {
+    case "Video":
```
ebefarooqui, line 30: It could be useful to make the media `types` constant strings (ie. `MEDIA_TYPE.IMAGE`, `MEDIA_TYPE.VIDEO`) so we can use those strs as single sources of truth

src/components/time/Timeline.js
```diff
@@ -339,12 +344,28 @@ class Timeline extends React.Component {
     return [null, null];
   }
 
+  onSelect(event) {
+    if (this.props.features.ZOOM_TO_TIMEFRAME_ON_TIMELINE_CLICK) {
+      const timeframe = Math.floor(
+        this.props.features.ZOOM_TO_TIMEFRAME_ON_TIMELINE_CLICK / 2
+      );
+      const start = d3.timeMinute.offset(event.datetime, -timeframe);
```
ebefarooqui, line 352: abstract `timeMinute` into a more general `timeInterval` so we can pass whichever interval set we want?

ebefarooqui (commented):
Looks good! Some minor edits from my side, but other than that, good to merge.

PR #226, opened 2022-03-22
fix: adds binary-searched event to list

msramalho:
the result of the binary search could be an event with a different id, so me must make sure to display it too
Another alternative, less explicit, is to do `let ptr = idx;` in the check forward, but that might not be so perceptive when looking at the code for the 1st time. 

detected here: https://github.com/bellingcat/ukraine-timemap/issues/9


ebefarooqui:
Feel free to merge. Looks good to me

breezykermo:
Thanks @msramalho!

PR #234, opened 2022-04-05
Feature/reduce bundle size

1cgonza:
- Switch from moment to dayjs, still supports multiple string formats but much smaller lib.
- Update d3 to 7.4.2 and use tree shaking to include only the parts needed.


breezykermo (approved):
Looks good, thanks! The d3 tree-shaking is long overdue :)

breezykermo:
@msramalho FYI, would you like a port of this for https://github.com/bellingcat/ukraine-timemap?

msramalho:
Yes, it's useful and not doing it also creates a cost of divergence.

Issue #24, opened 2018-11-29
Compact to one option for async events in config.js


breezykermo:
At the moment it is necessary to configure _two_ routes for events that are being loaded asynchronously: both `EVENT_EXT`, and `EVENT_DESC_ROOT`. It is not necessary to have both of these, as in our datasheet->timemap workflow, both routes are created from the same source data.

The redux/react architecture should be modified so that only _one_ option (`EVENT_EXT` to keep consistent with other async data) is required in 'config.js'.

breezykermo:
This will actually be nullified in #20, as `EVENT_DESC_ROOT` will become `EVENT_SOURCE_EXT` (or something similar), as it makes sense to separate the endpoint for the sources that compose an event, and retrieve that data on-demand, rather than loading it all up front.

PR #238, opened 2022-04-06
Feature/fix timeline shapes

1cgonza:
- Removes React warning of unique keys for Timeline shapes.
- `SQUARE` and `DIAMOND` need to be centered in the x axis, otherwise they are not consistent with the other shapes. 
- The `STAR` shape looks weird to me:
stars-deformed
With this PR it looks like this:
stars-fixed

### DIFF
--- a/src/common/utilities.js
+++ b/src/common/utilities.js
@@ -580,3 +580,5 @@ export function getFilterIdx(
 }
 
 export const isEmptyString = (s) => s.length === 0;
+
+export const isOdd = (num) => num % 2 !== 0;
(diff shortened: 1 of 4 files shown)


breezykermo:
Thanks for the PR @1cgonza. I believe that because of the way that `evShadows` potentially duplicates the return value of `renderShape` in the timeline Events.js logic, assigning the key as `event.id` is wrong and could lead to poor performance / internal React errors. See [this PR](https://github.com/bellingcat/ukraine-timemap/pull/24) for further details on this discussion.

@gaearon is working on a fix for these errors in this part of the codebase (and has better context than me in terms of what kinds of disaster duplicate IDs could invoke).

The other fixes around shape aesthetics are very welcome! If you are happy to remove the unique keys fix, I'll be happy to merge.

1cgonza:
Thank you @breezykermo, I was doing this in our map. Will keep track of the PR to better understand the issue and port the solution once it is resolved. 

I've updated the PR to revert those particular changes. Now it is only about the shapes.

PR #241, opened 2022-04-08
fix: downgrade react-error-overlay to `6.0.9` to fix HMR

fspoettel:
When working on the map, I noticed that hot module reloads are broken, resulting in the following error on each code change:

image

After that, the UI becomes unresponsive and I have to F5 the page.

I investigated it a bit and the root cause is a regression in a `react-dev-utils` component - see https://github.com/facebook/create-react-app/issues/11771 . Since we are using a lockfile, I downgraded to the previous patch version and HMR is working again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20618,9 +20618,9 @@
       }
     },
     "node_modules/react-error-overlay": {
-      "version": "6.0.10",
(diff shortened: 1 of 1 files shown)


breezykermo:
Fantastic, thank you - I noted this this morning and had scheduled some time in my calendar this afternoon to debug it. So very much appreciated!

PR #245, opened 2022-04-09
feat: serialize platform state to the url

### LINKED ISSUE #76: Serialize state of the platform as URL
In order to be able to cite the platform, the state at any given moment needs to be serialized and made to dynamically modify the URL. As a first step towards this, the selection of any event or set of events should be possible. Ultimately, it would be great to be able to cite certain moments in certain narratives as well.

fspoettel:
closes #76 

This is a proof-of-concept for serializing platform state to the URL.

* [x] events
* [x] time ranges
* [x] filters & colors

It's based on three components:
 
 1. `URLState` class that deals with serialization to the URL.
 2. Redux middleware that observes certain actions and updates URL state.
 3. Redux action that applies the serialized URL state on load.

I considered a few other approaches but ultimately settled on this for a few reasons: 
 1. introducing a `router` library would be a huge change and likely break downstream maps. this implementation is comparatively isolated.
 2. the `app` state is too big and complex to write it to the URL in its full form. we need to map it to a simpler representation.

In addition, I made the following design decisions:
 * decoupling `UrlState` from the middleware allows for behavior changes if we discover that we'd rather serialize state manually (e.g. with a share button).
 * `history.replaceState` is used istd. of `pushState`. When interacting with the map, a lot of different events might be accessed,  rendering the browser back button useless if the platform is linked to. `replaceState` seems preferable.

I would appreciate early feedback.


src/store/plugins/urlState.js
```diff
@@ -0,0 +1,75 @@
...
+export function urlStateMiddleware(store) {
+  return (next) => (action) => {
+    /**
+     * Selected events are persisted to the URL as a query parameter array -> `id=5&id=3`.
+     */
+    if (action.type === UPDATE_SELECTED) {
+      const state = new UrlState();
+      state.update(
+        "id",
+        action.selected.map(({ id }) => id)
```
fspoettel, line 68: this is a shortcut - we should probably use selectors here to read from `nextState` which is more robust:
```js
const nextState = next(action);
const selected = getSelectedIds(nextState);
// ...
return nextState;
```

fspoettel:
@breezykermo I would appreciate a high-level review of the PR in the current state - does this approach generally make sense to you? 

There a couple of `TODO`s for the current functionality that I'll address before merging.

The PR is missing support for `category` and `narrative` because I don't know how to configure and test these with the data I have. If you can point me to how to configure these, I'll cover them as well.

breezykermo:
Sure thing, I'll review this afternoon (EST)

src/store/plugins/urlState/urlState.js
```diff
@@ -0,0 +1,103 @@
...
+            .getAll(key)
+            .map((val) => this._decode(schema, val))
+        : this._decode(schema, this.url.searchParams.get(key));
+    });
+
+    return state;
+  }
+
+  serialize() {
+    window.history.replaceState(null, "", this.url);
```
breezykermo, line 86: I agree that `replaceState` is preferable here, as the primary (only?) use case for this is on first load, so the back button to whatever previous site the user was visiting shouldn't be displaced.

src/store/plugins/urlState/urlState.js
```diff
@@ -0,0 +1,103 @@
...
+  // Timeline
+  start_date: {
+    type: "date",
+    isArray: false,
+  },
+  end_date: {
+    type: "date",
+    isArray: false,
+  },
+});
```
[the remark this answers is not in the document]
fspoettel: I think we need two hooks per url key - one _select_ and one function to _apply_. The plugin could then look like this:
```js
const schema = {
  id: {
   // redux action that triggers `select`
    trigger: UPDATE_SELECTED,
   // select url state from current state. plugin then writes the URL
    select(state) {
      const ids = getSelected(state).map(({ id }) => id);
      return url.stringArray(ids);
    },
    // called from the `app` reducer hen rehydrating and maps URL state to state
    apply(appState, domain, ids) {
      appState.selected = urlState.id.map((id) => domain.events[id]);
    }
  }
}
```
what do you think?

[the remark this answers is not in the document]
breezykermo: Nice idea! Just to make sure I understand: `select` indicates the function from redux state to URL, whereas `apply` indicates from URL to redux state (vice versa), correct?

Let's make the function names and parameters clearer if this is correct. ie:
- `project_state_to_url(state, urlId)`, returns new URL as string
- `project_url_to_state(state, urlId, urlValue)`, returns new redux state (encompassing both `appState` and `domain`, which are subsections of `state`)

Following redux's philosophy, I suggest that these stay pure functions, qnd the necessary state and URL modifications are then done once that full intermediate representation has been generated. Thoughts?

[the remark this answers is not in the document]
fspoettel: sounds good, I'll give it a shot. 👍 

> (encompassing both appState and domain, which are subsections of state)

we would need a custom root reducer (something like this: https://stackoverflow.com/a/46039857) then but it's much clearer so I'd say it's worth it.

src/reducers/index.js
```diff
@@ -1,13 +1,26 @@
...
+
+export default decorateRootReducer(
+  rootReducer,
+  combineReducers({
+    app,
+    domain,
+    ui,
+    features,
+  })
+);
```
breezykermo, line 26: This is a nice workaround!

src/selectors/helpers.js
```diff
@@ -23,3 +25,25 @@ export function shuffle(a) {
   }
   return a;
 }
+
+export function toIds(arr, filterMapping) {
```
breezykermo, line 29: It's unclear to me from this name and argument what this function does. Can this be named more descriptively?

src/store/plugins/urlState/applyUrlState.js
```diff
@@ -0,0 +1,23 @@
...
+      console.error(err);
+    }
+  });
+
+  return nextState;
+}
+
+function isEmpty(o) {
+  return !Object.keys(o).length;
+}
```
breezykermo, line 23: minor: this should be a utility function imported

src/store/plugins/urlState/schema.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * Schema specifies how redux state maps to the url and vice versa.
+ * `trigger`: action that triggers a call to `dehydrate()`
+ * `type`: type of the mapped URL property
+ * `dehydrate()`: maps redux state to url state.
+ * `rehydrate()`:
+ *    maps url state to redux state.
+ *    !for performance reasons, this function works with a mutable ref to `state`!
```
breezykermo, line 40: Very clear explanation, thanks!

src/store/plugins/urlState/schema.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+export const SCHEMA = Object.freeze({
+  id: {
+    key: "id",
+    trigger: UPDATE_SELECTED,
+    type: SCHEMA_TYPES.NUMBER_ARRAY,
+    dehydrate(state) {
+      return getSelected(state).map(({ id }) => id);
+    },
+    // TODO: determine time range if `range` not set.
```
breezykermo, line 50: Is this TODO necessary before we can merge?

src/store/plugins/urlState/schema.js
breezykermo: is there any option to not use mutable ref to `state`, given that JS is pass by reference for objects and arrays? I just ask as I think that this final line might be a bit confusing / could use further explanation.

src/store/plugins/urlState/schema.js
```diff
@@ -0,0 +1,114 @@
...
+    },
+  },
+  color: {
+    key: "color",
+    trigger: UPDATE_COLORING_SET,
+    type: SCHEMA_TYPES.STRING_ARRAY,
+    dehydrate(state) {
+      return selectActiveColorSets(state);
+    },
+    // TODO: color parent if all children checked.
```
breezykermo, line 101: Likewise here: should this be completed?

src/store/plugins/urlState/schema.js
```diff
@@ -0,0 +1,114 @@
...
+      if (filter?.length) {
+        const filters = nextState.domain.associations.filter(
+          (x) => x.mode === ASSOCIATION_MODES.FILTER
+        );
+        const filterMapping = mapFilterIdsToPaths(filters);
+        nextState.app.associations.filters = filter.map(
+          (id) => filterMapping[id]
+        );
+      }
+    },
```
breezykermo, line 92: I like this format a lot, it's very clear to me what's happening and how to extend it.

breezykermo (commented):
Thanks again for this contribution @fspoettel, I'm excited to see this in action! I have put some preliminary comments inline, and I will look to test this feature with a few of our configurations at some point this week to test it out. Very excited to see it in action.

src/store/plugins/urlState/schema.js
fspoettel: We could return a change object here and `deepMerge` it in the reducer. This is more overhead though. Your call.

src/store/plugins/urlState/schema.js
fspoettel: No, I recorded smaller inprovements with TODO so that we can mege sooner.

fspoettel:
did you have a chance to test this yet @breezykermo?

breezykermo:
@fspoettel my apologies for the long delay- this looks really great! As I'm working on a small project with timemap in the coming days, I am going to merge this and incorporate it as part of the supported feature set going forward. If there are any issues I'll hotfix in that project.

PR #246, opened 2022-04-09
Update react-redux, redux and redux-thunk

### LINKED ISSUE #239: Update react-redux dependency
Now that `react` was updated to 18 (#233), it would make sense to also update `react-redux` to 7.

It seems to work OK when I bump it, but I am not very familiar with this codebase and haven't been using react-redux before, so I don't have a good intuition of what needs to be payed attention to when jumping a few major versions. It would be safer if somebody who does could have a look at this. 

Just bringing this up, because of a need for `useSelector` [here](https://github.com/bellingcat/ukraine-timemap/pull/33/files#diff-8bd7ccf93aeb8a9f6754bf104bcef088cd832d51f3cfa129c226351d449ea166R56-R65). Thanks.

jstolwijk:
Resolves  #239


breezykermo:
Thanks @jstolwijk! With what timemap configs have you tested this feature, if any? (I appreciate there aren't currently good instructions about this being necessary or even how to do so-- which I will open an issue for now-- so totally understandable if the answer is "None"!)

jstolwijk:
I tested the change by running the datasheet-server with the the default "dev" configuration.

PR #251, opened 2022-04-11
chore: replace `node-sass` with `sass`

### LINKED ISSUE #249: Use `sass` instead of `node-sass`
node-sass is no longer maintained, and so [dart sass](https://www.npmjs.com/package/sass) is the better package to use now.

fspoettel:
closes #249 

There are a couple of deprecation warnings in `video-react` pertaining to the upcoming removal of the `/` division syntax. See https://github.com/video-react/video-react/pull/424 We can ignore these for now

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "timemap",
       "version": "0.1.0",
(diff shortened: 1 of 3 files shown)


breezykermo:
Looks good to me, thanks!

PR #254, opened 2022-05-07
fix marker location collision bugs

dmitrigrabov:
Fixes two issues when two markers have same coordinates
1. When two markers have same coordinates, they will have same keys based on coordinates hash alone, which is best avoided in React. I've added parent event `id` to marker data to ensure generated keys are unique
2. Clicking on a marker, when there are two markers on same coordinates, was crashing the map. It looks like click was received by a `circle.event-hover` element, which gets removed from DOM immediately after. As a result Leaflet was throwing the following error `Cannot read properties of null (reading '_leaflet_disable_click') `

I've updated the behaviour to hide `circle.event-hover` elements which are not selected rather than remove them.


yarn.lock
```diff
@@ -0,0 +1,13452 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
fspoettel, line 1: this project uses `npm` (=> `package-lock.json`), so this file should be removed.

breezykermo:
Thank you @dmitrigrabov: I'm sorry for the long delay!

